PR #50, opened 2015-03-03
Make url

creationix:
This gives the `lit make` command url ability.  For example, the following are valid ways to build apps.

```sh
# get latest version from lit
lit make lit://luvit/luvit
# Specify a base version, gets best match
lit make lit://luvit/luvit@2.0.0
# shorthand for luvit:// prefix
lit make luvit/luvit
# Download zip from github and build from that           
lit make github://creationix/rye
# Download zip of luvi-up branch on github
lit make github://luvit/luvit@luvi-up
# Download arbitrary zip
lit make https://foo.bar.com/myapp.zip
# Make a shallow clone
lit make git://github.com/creationix/rye.git
```

### DIFF
--- a/commands/make.lua
+++ b/commands/make.lua
@@ -4,6 +4,10 @@ local pathJoin = require('luvi').path.join
 
 
 local cwd = uv.cwd()
-local app = args[2] and pathJoin(cwd, args[2]) or cwd
+local source = args[2] and pathJoin(cwd, args[2])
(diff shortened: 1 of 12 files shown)


rphillips:
Nice!

creationix:
The git urls shell out to `git` so it needs to be on the system to use them, but the github:// protocol (via zip url) is better when available anyway.

PR #53, opened 2015-03-13
ISSUE 52: Replaced " : " with " _ " for writing and reading keys to the filesystem, as Windows does not allow " : "

IZEDx:


rphillips:
We use this ```fingerprint:gsub("_", ":")``` two places... thinking we should write a function to sanitize the fingerprint

rphillips:
nevermind. it's a different translation.

Thank you for the fix.

creationix:
Looks good, but why rename keyPath and leave the old unused version.  It would be better to just edit keyPath directly.

creationix:
https://github.com/luvit/lit/commit/2b4d9fb112bb8d9612dcedfe27fe5e6a83fc9e44

PR #73, opened 2015-04-30
fix to semver to support buildnumber

rphillips:


deps/semver.lua
```diff
@@ -29,9 +30,9 @@ exports.normalize = normalize
 function exports.gte(first, second)
   if not second or first == second then return true end
   if not first then return false end
-  local a, b, c = parse(second)
-  local d, e, f = parse(first)
-  return (d > a) or (d == a and (e > b or (e == b and f > c)))
+  local a, b, c, x = parse(second)
+  local d, e, f, y = parse(first)
+  return (d > a) or (d == a and (e > b or (e == b and (f > c or (y > x)))))
```
creationix: I think this needs more, what about when `f == c` and `f < c`.  Shouldn't those be treated differently?

creationix: Maybe replace `or (y > x)` with `or (f == c and y > x)`

deps/semver.lua
```diff
@@ -41,10 +42,12 @@ assert(exports.gte("9.9.9", "9.9.9"))
 assert(exports.gte("9.9.10", "9.9.9"))
 assert(exports.gte("9.10.0", "9.9.99"))
 assert(exports.gte("10.0.0", "9.99.99"))
+assert(exports.gte("10.0.0-1", "10.0.0-0"))
 assert(not exports.gte(nil, "0.0.0"))
 assert(not exports.gte("9.9.9", "9.9.10"))
 assert(not exports.gte("9.9.99", "9.10.0"))
 assert(not exports.gte("9.99.99", "10.0.0"))
+assert(not exports.gte("10.0.0-0", "10.0.0-1"))
```
creationix: Add tests comparing `10.0.0-1` with `10.0.1-0`

creationix:
The match function also needs updating to support the 4th number.

creationix:
+1

rphillips:
Thanks Tim!

PR #75, opened 2015-05-01
Fix update on Windows not targetting .exe files

squeek502:
* Mirrors logic in libs/core.lua's realMake fn


creationix:
Oops, thanks.

Issue #72, opened 2015-04-29
Implement package search UI at luvit.io.


creationix:
A start has been prototyped and deployed, more data needs to be surfaced to make the search useful.

Also the API needs consistent data between search results and package indexing directly.

creationix:
See prototype at https://luvit.io/lit.html

james2doyle:
Looks good.

nihildeb:
nice! I know it's a prototype, but just sayin:
- if there are no results, say so. right now the querying... just disappears to a blank page
- links to homepages (maybe they are there, but I didn't see any)
- I can see that dep list getting out of hand (too long). it already sort of dominates the output. maybe a twisty to roll it up/down. It's not bad as it is though, just a thought...

nihildeb:
great! thanks @squeek502 

this is really helpful Tim, thanks!

nihildeb:
just noticed that the pluralize is not correct on the result count. When you get a single result it says:

> 1 matche

with multiple resuts it is correct, just adding the 's'

creationix:
The lit-browser.js interface still has several issues, but now thanks to the revamped search backend, we can do more interesting things.

- [ ] Show parsed search terms as little rectangles with close buttons.  When one is closed, start up search.
- [ ] Fix results count pluralization.
- [ ] Show message when there are no results suggesting broadening the search.
- [ ] Hyperlink results when possible, for example clicking on a dependency will do a name search for that name, clicking on a tag will search for all packages with that tag, author will search for all packages by that author, etc..

creationix:
All done and deployed to https://luvit.io/lit.html

Issue #56, opened 2015-03-24
Lit search package command


homer242:
It would be nice to have a lit search <package> command, wouldn't it ?

For now, it's pretty obscure for a new guy like me to have a look about packages available for luvit 2.0. So, I think a search command and a website like npm will be awesome to recruit new enthousiast for this tool :)

creationix:
Yep, I thought there was already a task for this.  I guess it's on the wrong repo.

creationix:
FWIW, you can browse the packages at lit.luvit.io if you have the jsonview extension in your browser.

james2doyle:
Could there maybe be a `http://lit.luvit.io/database` that returns the JSON tree of all packages? Or maybe a `http://lit.luvit.io/export` that triggers the DB export to a JSON file?

creationix:
The metadata itself isn't very easy to extract currently.  You need to walk the git graph and look for `package.lua` or `init.lua` for trees and look in the root blob for single-file packages.  Then you need a lua parser with a custom sandbox to extract the data.

I think I'll make a copy of the metadata on publish into the tag body as JSON.  Then it will be much easier to extract this data and perform queries.  Also I can document fields for things like homepage and docs so we can link to descriptions and READMEs for the packages.

creationix:
Depends on #67

creationix:
I just pushed an update to lit with a revamped search engine.

You can test it at https://luvit.io/lit.html

There are 4 kinds of filters (tag, author, name, and search), each can have
multiple terms that are ORed together.  The filters themselves are ANDed
together.

The `tag`, `author`, and `name` filters require a prefix like
`author:creationix` or `name:msgpack`.  If there is no prefix, it's assumed to
be a search term.

So `author:luvit` will show all packages by the luvit author.  But `author:luvit
stream` will show all packages by luvit *and* containing the search term
`stream`.  Prefixed terms are exact matches, use `*` to allow wildcards.  The
unprefixed search term is full-text search in name, keywords, description, and tags.

Also you can wrap terms in quotes to include spaces.  For example searching for
`"blue apple"` is one term `blue apple` while `blue apple` is two terms `blue`
and `apple`.

I don't like having both keywords and tags.  Should we encourage packages to
migrate towards "tags" or just use the "keywords" field internally for tag
searches?

There was one breaking change where the result "type" field used to be either
"package" or "author" and is now "directory" or "file" depending on if the
package is a single file or a while tree.

homer242:
very nice ! well done !

PR #76, opened 2015-05-06
Improve search and packages REST API

squeek502:
* Make /packages/{author}/{name}/{version} return similar metadata to search results (only difference is `type` and `url`). Note that /packages/{author}/{name}/{version} no longer returns `message` (the raw tag JSON).
 * Escape all special pattern characters in /search/{query} queries except * (allows a search like `coro-wrapper` to succeed)
 * Improve /search/{query} to allow for more robust searches:

`*`: All authors and packages
`*/`: All authors
`/*`: All packages (`*/*` would also work)
`author/`: All packages of the author (if the author name is an exact match)
`author/`: All authors that partially match the author query (if no exact matches were found)
`/package`: All packages that match the package query
`query`: All packages or authors that match the query

Misc changes
 * Routing no longer strips trailing slashes by sending the path through pathJoin. This was necessary to allow trailing slashes in search queries. Route patterns now take trailing slashes into account. This should cause no changes in functionality

Contributes towards #72

### DIFF
--- a/libs/api.lua
+++ b/libs/api.lua
@@ -84,9 +84,9 @@ GET /search/$QUERY -> packages/authors json {
 }
 
 Possible search queries:
-  ".*"        = All authors and packages
-  ".*/"       = All authors
-  "/.*"       = All packages
+  "*"        = All authors and packages
+  "*/"       = All authors
+  "/*"       = All packages
   "author/"   = All packages of the author (if the author name is an exact match)
   "author/"   = All authors that partially match the author query (if no exact matches were found)
   "/package"  = All packages that match the package query
@@ -200,8 +200,10 @@ return function (prefix)
       return next(authors) and authors
     end,
     "^/search/(.*)$", function (query)
-      -- escape all special pattern characters except . and *
-      query = query:gsub("[%(%)%%%+%-%?%[%]%^%$]", function(c) return "%" .. c end)
+      -- escape all special pattern characters except *
+      query = query:gsub("[%(%)%%%+%-%?%[%]%^%$%.]", function(c) return "%" .. c end)
+      -- expand * to .* for a glob-like wildcard
+      query = query:gsub("%*", ".*")
       local matches = {}
       local authorQuery, separator, packageQuery = query:match("^(.-)(/?)([^/]*)$")
       if authorQuery == "" then authorQuery = nil end
--- a/libs/api.lua
+++ b/libs/api.lua
@@ -34,7 +34,13 @@ GET /packages/$AUTHOR/$TAG/$VERSION -> tag json {
   object = "..."
   url = "/trees/..."
   type = "tree"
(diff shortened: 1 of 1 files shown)


libs/api.lua
```diff
@@ -136,29 +200,32 @@ return function (prefix)
       return next(authors) and authors
     end,
     "^/search/(.*)$", function (query)
+      -- escape all special pattern characters except . and *
+      query = query:gsub("[%(%)%%%+%-%?%[%]%^%$]", function(c) return "%" .. c end)
       local matches = {}
+      local authorQuery, separator, packageQuery = query:match("^(.-)(/?)([^/]*)$")
```
creationix: What about deep package names like `foo/bar/baz`?  I don't think any actually exist, but I've tried to keep that code path supported.

creationix: I guess this will work.  It puts the extra path segments in with the author.  That's probably the desired behavior.

For example, I might want to put the weblit-* packages under "creationix/weblit/*"

libs/api.lua
```diff
@@ -136,29 +200,32 @@ return function (prefix)
       return next(authors) and authors
     end,
     "^/search/(.*)$", function (query)
+      -- escape all special pattern characters except . and *
+      query = query:gsub("[%(%)%%%+%-%?%[%]%^%$]", function(c) return "%" .. c end)
```
creationix: If we're going to escape most lua patterns, then we should probably support using star globs.

Your lua pattern-fu seems better than mine, but if it helps, I started implementing glob matches on domains in weblit at https://github.com/creationix/weblit/blob/master/libs/weblit-app.lua#L224-L239

creationix:
Great work!  I'll have to consider the search semantics a bit more, but it's better than what I had already.

I just want to make sure we can also search on descriptions and tags eventually without changing the syntax again.

libs/api.lua
squeek502: Right, that convention follows how `lit-browser.js` would separate it (`author/author/../package`), which is what I used as a reference.

squeek502:
I wasn't sure how to handle keywords and descriptions. It might work out alright to just add keyword/description matching only when no `/` exists in the query, but that assumes keywords/descriptions wouldn't have `/` characters (is that a safe assumption?).

creationix:
I was thinking allowing parameters and having a little more advanced search
kinda like how gmail works.

"author:creationix tag:codec"

But any free-form search parameters would match against author, name,
keyword, or description.  We can give name highest preference and weight
them if want some basic relevance algorithm.

creationix:
Thanks for making this PR, but I ended up going a different route.  Let me know what you think of the recent changes to the search backend.

https://github.com/luvit/lit/issues/56#issuecomment-100028257

squeek502:
No worries, looks good.

PR #78, opened 2015-05-12
Compile Lua files when making luvi app

creationix:
This compiles all lua files when building a luvi app zip.  The final zip file isn't always smaller, but it uses a bit less ram.  Also authors can write verbose comments now without worrying about it hurting their luvi app.


rphillips:
+1

PR #84, opened 2015-06-05
RFR: Refactor

creationix:
Refactor the internals a lot to be more modular and saner logic.


lduboeuf:
@creationix oh i see, sorry for the confusion

libs/calculate-deps.lua
```diff
@@ -0,0 +1,78 @@
...
+            alias, meta.version, version)
+          log("version mismatch", message, "highlight")
+        end
+      end
+    else
+      local author, pname = name:match("^([^/]+)/(.*)$")
+      local match, hash = db.match(author, pname, version)
+
+      if not match then
+        error("No such "
```
rphillips: string.format here?

libs/core.lua
```diff
@@ -505,57 +307,76 @@ local function makeCore(config)
...
 
-
-  function core.make(path, target)
-    local fs, newPath = vfs(path)
-    local meta = pkg.query(fs, newPath)
-    if not meta then
-      error("Not a package at: " .. path)
+  local function defaultTarget(meta)
+    local target = meta.target or meta.name:match("[^/]+$")
+    if jit.os == "Windows" then
```
rphillips: almost thinking we should have an isWindows() function... i think this is the third check i saw

libs/core.lua
```diff
@@ -505,57 +307,76 @@ local function makeCore(config)
...
+  local function makeGit(target, url)
+    local path = (url:match("([^/]+).git$") or target or "app") .. ".git-clone"
+    log("downloading repo", url)
+    local stdout, stderr, code, signal = exec("git", "clone", "--depth=1", "--recursive", url, path)
+    if code == 0 and signal == 0 then
+      core.make(path, target)
+    else
+      error("Problem cloning: " .. stdout .. stderr)
+    end
+    exec("rm", "-rf", path)
```
rphillips: scary... maybe error out and tell the user the fix.

rphillips:
minor nit, we should probably add the luvit/luvi license to the source files as well.

libs/calculate-deps.lua
creationix: There isn't much static content, it would look like:

```lua
error(string.format("No such %s: %s%s",
  version and "version" or "package",
  name,
  version and '@' .. version or ''))
```

I'll change it if you prefer `string.format` here.

libs/core.lua
creationix: Or maybe an `isWindows` variable?  Not sure a function would be any simpler than a simple string check against a global property.

libs/core.lua
creationix: So don't `rm -rf` in the error case?  I'm pretty sure it's safe in the case of successful clone.

libs/calculate-deps.lua
rphillips: ah... yeah, that isn't much change. nm

libs/core.lua
rphillips: sure. isWindows variable at the top would be fine

libs/core.lua
rphillips: Hmm. If I was reading this source code, I would take a long pause on this line. maybe we could write a `rm -rf` function instead of shelling out.

creationix: Sure, I just realized that `rm` won't exist on windows anyway.  It's bad enough we're shelling out to git for cloning, but adding clone to my git implementation is a lot of work.

libs/core.lua
creationix: Actually, I'd rather not do this.  We already have a lot of local variables.  Too many cause performance issues in the jit vm.  The two checks for `jit.os` are not performance sensitive and it's pretty obvious what they do.

rphillips: ok. works for me

rphillips:
+1... i do think the license headers would be good to add as well

libs/core.lua
creationix: Ok, added `fs.rmrf(path)` that does a recursive rmdir/unlink.

PR #95, opened 2015-06-24
hexToBin/binToHex functionality already exists in openssl

DBarney:
the third parameter of the `openssl.digest.digest` function is a flag that can print out a binary string, or a hex string.

https://github.com/zhaozg/lua-openssl/blob/master/src/digest.c#L89


creationix:
That's handy.

creationix:
Is there nothing else using hex-bin?  I'm pretty sure it has several deps within lit.

DBarney:
You were right, there were a few last places that it was mentioned. I removed them and replaced them with `require('openssl').hex`

deps/websocket-codec.lua
```diff
@@ -1,8 +1,6 @@
 exports.name = "creationix/websocket-codec"
 exports.version = "1.0.5"
-exports.dependencies = {
-  "creationix/hex-bin@1.0.0"
-}
+exports.dependencies = {}
```
creationix: No need for an empty table, you can just remove this entry.

PR #96, opened 2015-06-25
item was nil when the connection closed. it would still reuse the socket

DBarney:
when the connection closes item would be nil, which would cause the for loop to terminate early. however the connection would still be put back into the reuse pool.


creationix:
Why did you change the for loop?  Wouldn't this be enough? https://github.com/luvit/lit/pull/97/files

PR #97, opened 2015-06-25
Account for closed sockets while reading body

creationix:


rphillips:
makes sense. I like the tweak so the sockets unref/ref themselves.

PR #100, opened 2015-07-10
add option to lit make to allow use of different luvi for cross platf…

sousoux:
…orm builds


sousoux:
I added another one into this PR. There is weirdness in the URI returned in the search API. As soon as the package API is hit the cache metadata uri is changed to the /tree URI. Personally I would prefer the tree URI to be the one returned all the time but it should definitely stay the same.

joerg-krause:
Really nice - unfortunatly not documented...

PR #103, opened 2015-07-10
RFR: Rewrite update system

### LINKED ISSUE #102: Update process needs improvement
- When lit is updating itself, it should only update itself first (not luvit or luvi)
 - Lit should use the correct (updated) luvi when building new versions of self.
 - Lit should use the logic in the zipball to build the zipball so that build code and source data match up.
 - After updating itself, it should start the new process and check for updates of luvit and luvi.

creationix:
Fixes #102


rphillips:
this should work nicely. thanks!

PR #107, opened 2015-07-14
fix(get-lit.ps1): Detect and work with 32bit archs on windows

### LINKED ISSUE #105: get-lit.ps1 on Windows 32bit fails.
This [1] needs some smarter detection: https://github.com/luvit/lit/blob/master/get-lit.ps1#L5

64bits:
```
C:> $ENV:PROCESSOR_ARCHITECTURE
IA64
```
32bits:
```
C:> $ENV:PROCESSOR_ARCHITECTURE
x86
```

kaustavha:
Fixes #105

### DIFF
--- a/get-lit.ps1
+++ b/get-lit.ps1
@@ -4,9 +4,12 @@ $LIT_VERSION = "2.1.8"
 
 if (test-path env:LUVI_ARCH) {
   $LUVI_ARCH = $env:LUVI_ARCH
-}
-else {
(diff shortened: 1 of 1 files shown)


rphillips:
Lgtm. Thanks!

creationix:
So if I open a cmd.exe environment from visual studio for 32-bit compiles, will this detect it as 32-bit on a 64-bit machine?

rphillips:
yes. this should be fine.

Issue #109, opened 2015-07-31
Implement shrinkwrap-like feature


creationix:
We need a way to grab the current set of deep dependencies and lock it into a `package.lua` file or add it to the lit protocol metadata and repository.  It might make sense to do this on publish for all packages.

When resolving dependencies, the higher items in the chain override the lower items.

creationix:
Implemented in #114

PR #114, opened 2015-08-03
Snapshot feature so published lit apps have reproducible builds

creationix:
This is a new behavior in lit.

When you add a new package to the database using `lit add` that contains dependencies and is a tree, lit will calculate the current set of dependencies for that package and save this snapshot hash along with the metadata stored in the git tag.

Then when using `lit make` with a `lit://foo/bar` url or using the `.zip` generator in the web API, the snapshot will be used to generate the zip file instead of the current set of dependencies.

This means that top-level apps will freeze all their deep dependencies at time of publish.  If an app wants to pick up big fixes in deep dependencies it will need to publish a new version.  A new build version would be enough.

But this also means that foo/bar/v1.2.3 will *always* have the same exact content and not change over time as libraries it depends on change.

### DIFF
--- a/libs/api.lua
+++ b/libs/api.lua
@@ -198,9 +198,15 @@ return function (db, prefix)
         error("Can only create zips from trees")
       end
 
-      local deps = {}
-      calculateDeps(db, deps, meta.dependencies)
-      hash = installDeps(db, hash, deps)
+      local hash
+      -- Use snapshot if there is one
+      if meta.snapshot then
+        hash = meta.snapshot
+      else
+        local deps = {}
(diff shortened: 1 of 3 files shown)


creationix:
I did find one possible downside with this new behaviour.  If you try to `lit add` a package who's dependencies don't exist, it will fail.  This extra strict behaviour is generally a good thing, but it breaks the lit unit tests since I'm not careful in the script to add all the packages in dependency order.

creationix:
This PR closes #109

creationix:
Actually, you can't snapshot single-file packages.  By fixing that, the unit tests pass again since most of lit's deps are single file.

PR #123, opened 2015-08-25
Hosts file

creationix:


creationix:
We shouldn't do this I think.  The libuv primitive of `uv.getaddrinfo` already takes this into account.

PR #126, opened 2015-08-30
Init command. Closes #124

### LINKED ISSUE #124: lit init
It would be cool if lit had an init command like npm to scaffold out a basic `package.lua` file or write to the head of a single `init.lua` file ala [luv-mime](https://github.com/creationix/luv-mime/blob/master/init.lua) or micro libraries like that.

james2doyle:
preview

I think I have got this down now. I'll explain some features:

* create either package.lua or init.lua
* defaults to your own username
* defaults to your own author details

This is my first PR that is Lua based, so let me know what needs to be fixed or changed before going ahead.

### DIFF
--- a/commands/init.lua
+++ b/commands/init.lua
@@ -5,6 +5,7 @@ local env = require("env")
 local log = require("log").log
 local pathJoin = require("luvi").path.join
 local cwd = require('uv').cwd()
+local sprintf = require("string").format
 
 local config = core.config
 
@@ -54,15 +55,6 @@ local function getConfig(name)
   end
 end
 
-local projectName = prompt("Project Name", config["username"] .. "/project-name")
-local projectVersion = prompt("Version", "0.0.1")
-local projectDescription = prompt("Description", "A simple description of my little package.")
-local projectTags = prompt("Tags (Comma Separated)", "lua, lit, luvit")
-local authorName = prompt("Author Name", getConfig("user.name"))
-local authorEmail = prompt("Author Email", getConfig("user.email"))
-local projectLicense = prompt("License", "MIT")
-local projectHomepage = prompt("Homepage", "https://github.com/" .. projectName)
-
 -- trim and wrap words in quotes
 function makeTags(csv)
   local tags = "{ "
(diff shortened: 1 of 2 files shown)


james2doyle:
Actually, I have a question already. What category in the [commands/README](https://github.com/luvit/lit/blob/master/commands/README) do I add the help for this?

commands/init.lua
```diff
@@ -0,0 +1,100 @@
...
+  -- trim trailing comma and space
+  tags = tags:sub(0, (#tags - 2))
+  return tags .. " }"
+end
+
+local data = ""
+
+-- manually craft strings to control formatting
+if output == "init.lua" then
+  data = "exports.name = \"" .. projectName .. "\"\nexports.version = \"" .. projectVersion .. "\"\nexports.dependencies = {}\nexports.description = \"" .. projectDescription .. "\"\nexports.tags = " .. makeTags(projectTags) .. "\nexports.license = \"" .. projectLicense .. "\"\nexports.author = {\n  name = \"" .. authorName .. "\",\n  email = \"" .. authorEmail .. "\"\n}\nexports.homepage = \"" .. projectHomepage .. "\"\n"
```
rphillips: Hi James! Thank you for working on this. This could be refactored to be something like:
```lua
local t = {
  name = projectName,
  version = projectVersion,
  dependencies = {},
  tags = makeTags(projectTags),
  etc... }

for k, v in pairs(t) do
  if type(v) == "string" then
    print(string.format("exports.%s = \"%s", k, v))
  if type(v) == "table" then
    print(string.format("exports.%s = {}, k))
  end
end
```

Alternatively, we could use a multi line template:
```
tmpl = [[
exports.name = "%s"
exports.version = "%s"
exports.dependencies = {}
]]
```
then pass it to string.format()

This feature will be super useful. personally, I like the template format.

rphillips:
I like the section `Package Management`, but Tim may have a different opinion.

james2doyle:
Oh man! I was trying to find a better way but didn't think of the string
format template. It's a no brainer now. I'll switch to the string format.

I'll get to this ASAP

creationix:
On Aug 30, 2015 3:01 PM, "Ryan Phillips" <notifications@github.com> wrote:
>
> I like the section Package Management, but Tim may have a different
opinion.

Yep, sounds great.

creationix:
My only suggestion is to leave the quotes out of the template and write a
simple quoting function to make it a little more robust. Though, we don't
need to support all input strings I guess, soo maybe that's less needed
here.

james2doyle:
Ok @creationix I think by using the `%q` format over `%s` will fix the quoting issue. This is a simple test:

quote example

And here is the description from the Lua.org docs:

> To quote an arbitrary string in a secure way, the format function, from the standard string library, offers the option "%q". It surrounds the string with double quotes and properly escapes double quotes, newlines, and some other characters inside the string.

I didn't realize that the `%q` option existed. No wonder I couldn't find any good Lua string escape functions, they aren't really needed for quotes and string literals.

It doesn't escape `$` or backticks, is that a problem here?

creationix:
Wow, that's great, I never knew about `%q`.  That should work perfectly.

james2doyle:
I pulled the latest lit and merge it into this branch so it's good to go

creationix:
Lovely work!

squeek502:
Very cool. :+1:

creationix:
This is in the fresh lit v2.2.13 version.  `lit update`!

PR #131, opened 2015-10-28
Action names

james2doyle:
The `action` variable is now a table with 2 words for each case.

Does this work?


creationix:
Thanks.

PR #135, opened 2015-11-04
Fix .gitconfig not being found on Windows when cwd is not on the home drive

squeek502:
* Use %HOMEDRIVE% env var to get the correct drive, as %HOMEPATH% only contains the drive-relative path.
 * Example: %HOMEDRIVE% contains "C:", %HOMEPATH% contains "\Users\username". Putting them together gets you to full absolute path
 * More info: http://stackoverflow.com/a/2115116

### DIFF
--- a/commands/auth.lua
+++ b/commands/auth.lua
@@ -24,7 +24,7 @@ local function confirm(name, value)
   return value
 end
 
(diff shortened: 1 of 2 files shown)


creationix:
Nice work.

PR #137, opened 2015-11-11
git clone should be recursive so that the git submodule is pulled down.

DBarney:
Uniaika from the IRC was having an issue where lit didn't have the git submodule. The instructions don't quite work as they are.


creationix:
Thanks!

PR #138, opened 2015-11-11
How to compile with a custom luvi build.

Annwenn:


creationix:
Thanks for helping out!

PR #155, opened 2016-01-05
RFR: Switch to luvit-loader

creationix:
In order to make the luvit ecosystem more friendly with the existing lua ecosystem, luvit's custom require is being replaced with a loader plugin for the native require.  This PR ports lit to this new system.

 - Added `luvit-loader.lua` to the root.  This adds a new loader that implements the old luvit require search paths as a loader function.  The main difference is files are loaded using plain `loadfile` or `loadstring` without setting a custom environment and injecting fake globals.

- Lit metadata for single-file modules is now in a special comment instead of setting properties on an assumed global `exports` table.

- Exports value is always returned since there is no longer a global `exports` table.

- Lit's auto-config, auto-core, and command systems were changed slightly to never yield during a require call since we're using real require now.

This depends on changes to luvi (removing the `luvit/require` hook) and changes to the git submodule.  Look for the `luvly` branches of both those projects.

To test this, make sure to use a custom luvi with the proposed changes from it's `luvly` branch.


creationix:
Still left to do.

 - Add support in lit to read the new metadata format.
 - Add fallback support in loader to look in bundle if there is a bundle and file wasn't found on disk.
 - Add support in loader to load binary modules.
 - Add module caching in loader.  I don't think lua does this for us.
 - Use realpath in libuv to resolve paths for better caching.

creationix:
Updated luvi is being published.  Most the TODO items regarding luvit-loader functionality can't be tested in lit since it doesn't use those features.  I'll punt them for the next step integrating with luvit.

rphillips:
+1

Issue #161, opened 2016-01-18
Error while running `lit make lit://luvit/luvit` on Windows.


xpol:
F:\luvit>lit make lit://luvit/luvit
lit version: 3.0.0
luvi version: v2.5.2
command: make lit://luvit/luvit
fail: [string "bundle:libs/pkg.lua"]:30: module 'git' not found:
        no file 'bundle:libs/deps/git'
        no file 'bundle:libs/deps/git.lua'
        no file 'bundle:libs/deps/git/init.lua'
        no file 'bundle:libs/libs/git'
        no file 'bundle:libs/libs/git.lua'
        no file 'bundle:libs/libs/git/init.lua'
        no file 'bundle:deps/git'
        no file 'bundle:deps/git.lua'
        no file 'bundle:deps/git/init.lua'
        no file 'bundle:libs/git'
        no file 'bundle:libs/git.lua'
        no file 'bundle:libs/git/init.lua'
        no field package.preload['git']
        no file '.\git.lua'
        no file 'F:\luvit\lua\git.lua'
        no file 'F:\luvit\lua\git\init.lua'
        no file 'C:\Lua\lua-5.1.5\x86\LuaRocks\2.2\lua\git.lua'
        no file 'C:\Lua\lua-5.1.5\x86\LuaRocks\2.2\lua\git\init.lua'
        no file 'C:\Lua\lua-5.1.5\x86\LuaRocks\systree\share\lua\5.1\git.lua'
        no file 'C:\Lua\lua-5.1.5\x86\LuaRocks\systree\share\lua\5.1\git\init.lua'
        no file '.\git.dll'
        no file 'F:\luvit\git.dll'
        no file 'F:\luvit\loadall.dll'
        no file 'C:\Lua\lua-5.1.5\x86\LuaRocks\systree\lib\lua\5.1\git.dll'
stack traceback:
        [C]: in function 'require'
        [string "bundle:libs/pkg.lua"]:30: in main chunk
        [string "bundle:luvit-loader.lua"]:200: in function <[string "bundle:luvit-loader.lua"]:195>
        [C]: in function 'require'
        [string "bundle:libs/core.lua"]:34: in main chunk
        [string "bundle:luvit-loader.lua"]:200: in function <[string "bundle:luvit-loader.lua"]:195>
        [C]: in function 'require'
        [string "bundle:commands/make.lua"]:2: in function <[string "bundle:commands/make.lua"]:1>
[...]

PR #162, opened 2016-01-18
API: Add /packages/$AUTHOR/$TAG/latest and /packages/$AUTHOR/$TAG/latest.zip

squeek502:
- Gets the metadata or the zip bundle of the latest version of the package
 - See #161

---

No clue if this is the optimal solution, but I figured I'd offer it as a PR just in case. Note also this makes the version parameter in the API a bit more flexible: both `/packages/author/name/v1.0.0` and `/packages/author/name/1.0.0` will work now (whereas before, the preceding `v` was required).

### DIFF
--- a/libs/api.lua
+++ b/libs/api.lua
@@ -63,7 +63,9 @@ GET /packages/$AUTHOR/$TAG/$VERSION -> tag json {
   }
   message = "..."
 }
+GET /packages/$AUTHOR/$TAG/latest -> tag json of the most recent version
 GET /packages/$AUTHOR/$TAG/$VERSION.zip -> zip bundle of app and dependencies
+GET /packages/$AUTHOR/$TAG/latest.zip -> zip bundle of the most recent version
 
 GET /search/$query -> list of matches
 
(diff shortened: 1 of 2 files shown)


libs/api.lua
```diff
@@ -192,8 +194,12 @@ return function (db, prefix)
         search = prefix .. "/search/{query}",
       }
     end,
-    "^/packages/([^/]+)/(.+)/v([^/]+)%.zip$", function (author, name, version)
-
+    "^/packages/([^/]+)/(.+)/([^/]+)%.zip$", function (author, name, version)
+      if version == "latest" then
+        version = (db.offlineMatch or db.match)(author, name)
+      else
+        version = version:sub(1,1) == "v" and version:sub(2) or version
```
creationix: since we already have an `else` block here, should probably use:

```lua
elseif version:sub(1,1) == "v" then
  version = version:sub(2)
```
... so as to not assign version to itself in the common case.

creationix:
Just a tweak to the logic style, otherwise looks good.

squeek502:
Made that change.

creationix:
@squeek502 thanks for all the support.  Sorry if my responses are sometimes slow.  I appreciate all the help you give.

squeek502:
No problem! You do a great job maintaining this project; I'm just happy to help where I can.

Issue #173, opened 2016-02-15
Add tls support for luv streams.


creationix:
Depends on https://github.com/luvit/luv/issues/219 and a way to create contexts (luaossl is a good option).

PR #174, opened 2016-02-24
Switch to secure-socket instead of coro-tls

creationix:
This enabled coro-net to handle tls as an option so that encrypted
streams can use both tls and codecs at the same time.

Also it gets us closer to handling ssl/tls at the C level.

(related to #173)


creationix:
All the lit tests pass.  I'm testing out the new coro-net changes and secure-socket library a bit more in my new websocket stuff.

rphillips:
nice, +1

PR #178, opened 2016-03-17
Add memory and file descriptor metrics to Lit server.

sam-falvo:
To access, hit the /metrics endpoint of the Lit server.  This will produce a
plain text, machine-readable result informing the user of the total memory
consumed by the Lit server, as well as the total number of file descriptors
used by Lit at the time the metrics were requested.

### DIFF
--- a/libs/api.lua
+++ b/libs/api.lua
@@ -149,7 +149,6 @@ local function collectStats()
 
     local iter = fs.scandir(path)
     for entry in iter do
-      print(entry.name)
       entries = entries + 1
     end
     return entries
--- a/libs/api.lua
+++ b/libs/api.lua
@@ -81,6 +81,7 @@ local exportZip = require('export-zip')
 local calculateDeps = require('calculate-deps')
 local queryDb = require('pkg').queryDb
 local installDeps = require('install-deps').toDb
+local ffi = require('ffi')
 
 local litVersion = "Lit " .. require('../package').version
 
@@ -130,6 +131,44 @@ end
 
 local metaCache = {}
 
+-- Collect statistics about this server's running instance.
+-- Return these metrics in a table.  The intended consumer of this
+-- information should render it in JSON format for transmission to
+-- a time-series database and/or visualization tools.
+local function collectStats()
+  local entry, fdsUsed = 0, 0
+  local memoryUsed = 1024 * collectgarbage("count")
+
+  local function countFDs(path)
+    local entries = 0
+
+    -- You might think this leaks a filehandle, but it actually doesn't.
+    -- You can confirm this yourself by hitting /metrics several times in a row,
+    -- and observing that lua.fds.total does not increase with each GET request.
+
+    for entry in io.popen([[ls -1 "]]..path..[["]]):lines() do
+      entries = entries + 1
+    end
+    return entries
+  end
+
+  local howToDetermineFDs = {
+    Linux = function() return countFDs("/proc/self/fd") end,
+    OSX = function() return countFDs("/dev/fd") end,
+    Other = function() return -1 end
+  }
+
+  local os = ffi.os
+  if howToDetermineFDs[os] == nil then
+    os = "Other"
+  end
+
+  return {
+    ["lua.mem.used"] = memoryUsed,
+    ["lua.fds.used"] = howToDetermineFDs[os]()
+  }
+end
+
 return function (db, prefix)
 
   local function makeUrl(kind, hash, filename)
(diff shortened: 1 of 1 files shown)


creationix:
Looks great.  Is there any particular reason you're using plain-text?  All the other lit endpoints return JSON.

sam-falvo:
Ease of getting something working.  I'll send a second patch that makes it emit JSON.  Is there a recommended JSON serialization library I should look into?  Or is hard-wired JSON sufficient?

creationix:
There is no JSON in lua, but luvit comes with one.  

```lua
local jsonEncode = require('json').encode
local json = jsonEncode(somevalue)
```

squeek502:
Also note that if a route returns a table, it will get auto-converted into JSON.

creationix:
Thanks @squeek502, I forgot about that.

sam-falvo:
Thanks; I'll make that change in a few minutes.  Stand by!

sam-falvo:
Done!

creationix:
+1

libs/api.lua
```diff
@@ -130,6 +131,44 @@ end
...
+  local memoryUsed = 1024 * collectgarbage("count")
+
+  local function countFDs(path)
+    local entries = 0
+
+    -- You might think this leaks a filehandle, but it actually doesn't.
+    -- You can confirm this yourself by hitting /metrics several times in a row,
+    -- and observing that lua.fds.total does not increase with each GET request.
+
+    for entry in io.popen([[ls -1 "]]..path..[["]]):lines() do
```
creationix: I'm pretty sure this is a blocking operation which will cause problems for the server's event loop.  Would be better to use the luvit APIs for traversing a directory?

creationix: Since the rest of lit is based on coroutines, let's use coro-fs here. For in a sample program you would do:

```lua
local fs = require('coro-fs')

-- Create and run coroutine so that we can use coroutine-based APIs
-- This function will only block till the first time the coroutine suspends itself.
coroutine.create(function ()
  for entry in fs.scandir do
    print(entry.name, entry.type)
  end
end)()

```

libs/api.lua
```diff
@@ -130,6 +131,44 @@ end
...
+    Other = function() return -1 end
+  }
+
+  local os = ffi.os
+  if howToDetermineFDs[os] == nil then
+    os = "Other"
+  end
+
+  return {
+    ["lua.mem.used"] = memoryUsed,
```
creationix: Could we add some more entries from libuv's functions?  The 3 memory related functions might be useful.

See this task Adam is working on for a listing of the more interesting functions. https://github.com/virgo-agent-toolkit/super-agent/issues/32

libs/api.lua
creationix: Here is a better example https://github.com/luvit/lit/blob/b3590f4bf5d4e271126f8ed2b6f66462c9fbe21d/libs/get-installed.lua#L25-L27

sam-falvo:
Now with more asynchronous flavor and wholesome concurrency!

libs/api.lua
```diff
@@ -130,6 +132,45 @@ end
...
+  local memoryUsed = 1024 * collectgarbage("count")
+
+  local function countFDs(path)
+    local entries = 0
+
+    -- You might think this leaks a filehandle, but it actually doesn't.
+    -- You can confirm this yourself by hitting /metrics several times in a row,
+    -- and observing that lua.fds.total does not increase with each GET request.
+
+    local iter = fs.scandir(path)
```
creationix: Just a note, that is path doesn exist (ENOENT), then iter will be nil.  You can wrap fs.scandir in an assert if you want it to just throw in this case.

sam-falvo: Those directories should always exist; they're exposed by the kernel.

libs/api.lua
sam-falvo: When they're ready for consumption, sure.  But, not sure what value they provide right at the moment.  RSS might be useful, but I suspect it'll track `memoryUsed` above closely.  Memory used could be useful to distinguish memory consumed by libuv and Lua, I guess.

creationix:
Should probably add self documenting entry in the TOC result at the `/` route.

libs/api.lua
```diff
@@ -168,6 +209,9 @@ return function (db, prefix)
   end
 
   local routes = {
+    "^/metrics$", function (hash, path)
```
creationix: no need for the `hash` and `path` arguments here, there are no pattern matches in this route.

creationix:
Since this is already closed I just tweaked it myself https://github.com/luvit/lit/commit/b07a7f02480ec824dda5a9e3f0c6fd0840685766

PR #182, opened 2016-04-12
Better error message when package name is malformed

squeek502:
See luvit/luvit#882

Will now give the following error instead:

```
fail: [string "bundle:libs/core.lua"]:133: Malformed package metadata. Package name must be of the format 'author/package'.
```

Note that malformed (or missing) version errors are already handled by semver.normalize within pkg.normalize.

### DIFF
--- a/libs/core.lua
+++ b/libs/core.lua
@@ -129,7 +129,7 @@ local function makeCore(config)
       error("Not a package: " .. path)
     end
     local author, name, version = pkg.normalize(meta)
-    if not author or not name then
+    if not (author and name) then
       error("Malformed package metadata. Package name must be of the format 'author/package'.")
     end
     if config.upstream then core.sync(author, name) end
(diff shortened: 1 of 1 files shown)


libs/core.lua
```diff
@@ -129,6 +129,9 @@ local function makeCore(config)
       error("Not a package: " .. path)
     end
     local author, name, version = pkg.normalize(meta)
+    if not author or not name then
```
creationix: Change to `not (author and name)` to save an opcode.  Unless luajit already does this internally.

creationix: Or even if it does optimize this, changing the logic seems easier to me to understand the intent.  What do you think?

creationix:
Thanks, looks great, just one nit about logic style.

squeek502:
Made that change.

PR #188, opened 2016-06-24
Fix local deps not being used consistently

### LINKED ISSUE #184: Be consistent using local deps
As discovered in another issue <https://github.com/luvit/luvit/pull/892#issuecomment-217502484>, lit doesn't include local deps always on certain platforms.

squeek502:
Closes #184. Tested and working, see https://travis-ci.org/squeek502/luvit/jobs/140012772 (no dependencies fetched from remote)

f57e2a4 is needed because the lit deps were also being fetched on Travis and locally on my Windows machine. Here's an example build without that fix included: https://travis-ci.org/squeek502/luvit/jobs/140010976

Let me know if the solutions can be improved.

### DIFF
--- a/libs/get-installed.lua
+++ b/libs/get-installed.lua
@@ -26,6 +26,10 @@ return function (fs, rootPath)
     if not iter then return end
     for entry in iter do
       local baseName
+      if not entry.type then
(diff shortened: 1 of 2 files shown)


creationix:
Nice work!

PR #189, opened 2016-06-27
Bump lit version in get-lit scripts

squeek502:
See luvit/luvit#916


creationix:
Silly me.  I was thinking the get-lit scripts were in the luvit repo for some reason.  Sorry for forgetting to update these then publishing the new lit version.

PR #191, opened 2016-08-19
Some luvit-loader fixes/improvements

squeek502:
Fixes for some some issues I ran into while making [luver](https://github.com/squeek502/luver).

Let me know if you'd like any more information about any of the changes or if you see a better way to handle any of these issues.

### DIFF
--- a/luvit-loader.lua
+++ b/luvit-loader.lua
@@ -197,7 +197,7 @@ local function loader(dir, path, bundleOnly)
         return package.loaded[key]
       end
       local code = bundle.readfile(fullPath)
-      local module = loadstring(code, key)()
+      local module = loadstring(code, key)(key)
       package.loaded[key] = module
       return module
     end, key
@@ -207,7 +207,7 @@ local function loader(dir, path, bundleOnly)
     if package.loaded[fullPath] then
       return package.loaded[fullPath]
     end
-    local module = assert(loadfile(fullPath))()
+    local module = assert(loadfile(fullPath))(fullPath)
     package.loaded[fullPath] = module
     return module
   end
--- a/luvit-loader.lua
+++ b/luvit-loader.lua
@@ -222,7 +222,13 @@ table.insert(package.loaders, 1, function (path)
     return
   end
 
-  local caller = debug.getinfo(3, "S").source
+  local level, caller = 3
+  -- Loop past any C functions to get to the real caller
+  -- This avoids pcall(require, "path") getting "=C" as the source
+  repeat
+    caller = debug.getinfo(level, "S").source
+    level = level + 1
+  until caller ~= "=[C]"
   if string.sub(caller, 1, 1) == "@" then
     return loader(pathJoin(cwd, caller:sub(2), ".."), path)
   elseif string.sub(caller, 1, 7) == "bundle:" then
--- a/luvit-loader.lua
+++ b/luvit-loader.lua
@@ -183,7 +183,7 @@ local function loader(dir, path, bundleOnly)
          try(pathJoin(dir, "libs", path)) then
         break
       end
-      if #dir < 2 then
+      if dir == pathJoin(dir, "..") then
         return table.concat(errors)
       end
       dir = pathJoin(dir, "..")


luvit-loader.lua
```diff
@@ -183,7 +183,7 @@ local function loader(dir, path, bundleOnly)
          try(pathJoin(dir, "libs", path)) then
         break
       end
-      if #dir < 2 then
+      if dir == pathJoin(dir, "..") then
```
creationix: Interesting, what issue is this working around or fixing?

luvit-loader.lua
```diff
@@ -197,7 +197,7 @@ local function loader(dir, path, bundleOnly)
         return package.loaded[key]
       end
       local code = bundle.readfile(fullPath)
-      local module = loadstring(code, key)()
+      local module = loadstring(code, key)(key)
```
creationix: Good idea, we should also include `@` in the path if it's a real path on the filesystem, then tools like luacov can work.

luvit-loader.lua
```diff
@@ -207,7 +207,7 @@ local function loader(dir, path, bundleOnly)
     if package.loaded[fullPath] then
       return package.loaded[fullPath]
     end
-    local module = assert(loadfile(fullPath))()
+    local module = assert(loadfile(fullPath))(fullPath)
```
creationix: This should be `'@'..fullPath` for reasons stated above.

luvit-loader.lua
creationix: So even though this is a bundle path, we can sometimes get a real system path (like when luvi is being run against a directory on the disk).  See how this is done in luvit's require. https://github.com/luvit/luvit/blob/1e2b4593c5552a642f8bf31142f440848d8b226d/deps/require.lua#L292-L297

luvit-loader.lua
squeek502: See [this commit message](https://github.com/luvit/lit/pull/191/commits/0c2bdc984a93d499417b5c8109e2cf9815eb21df):

> Fix infinite loop when a module is not found on Windows
>  - Windows root is 2 characters, so the #dir < 2 check would fail over and over (pathJoin("C:", "..") = "C:")

luvit-loader.lua
squeek502: Is there precedent for that convention being used in package.loaded? It seems like whatever key is used for package.loaded is what should be passed to the function, so if '@' is appended to this parameter then it should also be appended to the package.loaded key.

Basically, I think this should always succeed:

```
-- module.lua
return {name = ...}
```

```
local module = require('module')
assert(module == package.loaded[module.name])
```

squeek502:
Another thing I just realized: require automatically adds a key to package.loaded using the module name that is passed in, which can cause some weirdness. For example:

```
-- /usr/main.lua
local r = require("./relative.lua")
-- luvit-loader will set package.loaded['/usr/relative.lua'] and require will set package.loaded['./relative.lua'] to the return value of the loaded file
assert(package.loaded['./relative.lua'] == package.loaded['/usr/relative.lua'])
```

```
-- /usr/sub/test.lua
local r = require("./relative.lua")
-- r will be the cached value of package.loaded['./relative.lua'], not the return value of loading /usr/sub/relative.lua
```

luvit-loader.lua
squeek502: Seems the '@' prepend should go on the second `loadstring` parameter, not passed to the loaded function. Something like:

```
local path = pathJoin(bundle.base, "./" .. fullPath)
local module = loadstring(code, path)(key)
```

?

creationix: Right, it should only be added to the loadstring call for the filename.

luvit-loader.lua
creationix: I've never heard of this.  If it's harmless then it's fine, but they seem like different typed of names to me.

creationix:
Yes, the package.loaded insertion weirdness is why I insert this loader at the front of the list, even before the one that looks in package.loaded.

squeek502:
As far as I can tell, that lookup is done in require itself, not in a loader, see https://github.com/lua/lua/blob/5.1.x/src/loadlib.c#L454-L461

luvit-loader.lua
squeek502: It allows usage of the [module](http://lua-users.org/wiki/AlternativeModuleDefinitions) function like so:

```lua
module(...)
```

`module` is deprecated/removed in Lua > 5.1, but Lua still passes the module name as a parameter to loaded files. In [Lua > 5.1, the filename is also passed as the second parameter to the loaded module](https://github.com/lua/lua/blob/5.2.x/src/loadlib.c#L400).

luvit-loader.lua
squeek502: I believe that change will require a few other changes as well, as the `bundle:` prefix is what is checked using debug.getinfo() in the loader. I haven't done much with the bundle side of the loader, so I'll have to look into that more and write some test cases for it.

squeek502: Looks like I was right--appending '@' to the path results in module not found errors when lit is built (when using the `bundle:` prefix, this command works fine):

```
lit version: 3.4.3
luvi version: v2.7.4
command: ls
fail: C:\Users\Ryan\Programming\Luvit\lit.exe\deps\git\init.lua:2: module './core' not found:
```

Making bundles use the '@' prefix will require some restructuring of how `luvit-loader` checks the caller and decides whether or not to set the `bundleOnly` flag.

creationix:
That's what I thought as well, but when I was testing it, I did find that by inserting my loader first, it somehow got around the problem.  It's been a while so I may be remembering something wrong.

creationix:
Did you still want to merge this in.  Sorry I got distracted with travels.

squeek502:
Yeah, I think this is good to go. The '@' prefix when loading bundled dependencies would [require a larger rewrite](https://github.com/luvit/lit/pull/191#r75589020) and potentially require changes to luvi (since [it also uses the 'bundle:' prefix](https://github.com/luvit/luvi/blob/master/src/lua/luvibundle.lua#L321)).

PR #194, opened 2016-09-30
expose syntax errors in package.lua eval

james2doyle:
Similar to #193 but with the exposed errors from the `evalModule` function.

#### Example output with missing comma

```
lit version: 3.4.3
luvi version: v2.7.4
command: install
load config: /Users/james/.litconfig
fail: [string "bundle:libs/core.lua"]:529: [string "/Users/james/Sites/_git/lit/tests/bad-package/"]:6: '}' expected (to close '{' at line 4) near '"creationix/weblit-auto-headers@2.0.2"'
stack traceback:
	[C]: in function 'error'
	[string "bundle:libs/core.lua"]:529: in function 'installDeps'
	[string "bundle:commands/install.lua"]:6: in function <[string "bundle:commands/install.lua"]:1>
	[string "bundle:main.lua"]:52: in function <[string "bundle:main.lua"]:39>
	[C]: in function 'xpcall'
	[string "bundle:main.lua"]:39: in function <[string "bundle:main.lua"]:31>
```


libs/core.lua
```diff
@@ -523,7 +523,12 @@ local function makeCore(config)
   end
 
   function core.installDeps(path)
-    local meta = pkg.query(gfs, path)
+    local meta, err = pkg.query(gfs, path)
+    -- meta will be `nil` if package is malformed
+    if not meta then
+      error(err)
+      return
+    end
```
creationix: Since you're just throwing the error a simple `local meta = assert(pkg.query(gfs, path))` would have the same effect and be a lot clearer.

james2doyle: Nice. I made the change

Issue #200, opened 2017-02-01
#197 breaks coro-http


SinisterRectus:
Fresh installation:
```
sinister@ubuntu:~/new$ ./lit install creationix/coro-http
lit version: 3.5.2
luvi version: v2.7.6
command: install creationix/coro-http
load config: /home/sinister/.litconfig
connecting: wss://lit.luvit.io/
fetching: 1 object
fetching: 1 object
fetching: 1 object
fetching: 1 object
fetching: 1 object
including dependency: coro-channel (3.0.0)
including dependency: coro-http (2.2.0)
including dependency: coro-net (3.0.0)
including dependency: coro-wrapper (3.0.0)
including dependency: http-codec (3.0.1)
installing package: creationix/coro-wrapper@v3.0.0
installing package: creationix/coro-channel@v3.0.0
installing package: creationix/coro-http@v2.2.0
installing package: creationix/coro-net@v3.0.0
installing package: luvit/http-codec@v3.0.1
done: success
```

Reduced test code:
```lua
local http = require('coro-http')

coroutine.wrap(function()
	local res, data = http.request('GET', 'http://google.com/')
	p(res)
	p(data)
end)()
```

Error:
```
sinister@ubuntu:~/new$ ./luvit test.lua
Uncaught Error: /home/sinister/new/deps/coro-channel.lua:62: /home/sinister/new/deps/coro-wrapper.lua:109: index must be a number if set
stack traceback:
        [C]: in function 'assert'
        /home/sinister/new/deps/coro-channel.lua:62: in function </home/sinister/new/deps/coro-channel.lua:54>
        [C]: in function 'run'
        [string "bundle:init.lua"]:52: in function <[string "bundle:init.lua"]:47>
        [C]: in function 'xpcall'
        [string "bundle:init.lua"]:47: in function 'fn'
        [string "bundle:deps/require.lua"]:309: in function <[string "bundle:deps/require.lua"]:265>

```

squeek502:
As an aside, it'd also be nice to get some tests written for the `coro-` packages.

creationix:
Yes, some dedicated tests and docs for the coro-* libs would be awesome.

I was testing this against the lit unit tests which I thought would be sufficient.  I was wrong.

squeek502:
Found the issue: `http-codec` from the Luvit bundle is being used instead of the local `deps/http-codec`.

creationix:
Oh, I do remember seeing this, but that was before I got sick and checked out.

[...]

PR #202, opened 2017-02-03
Tests for coro- libraries

squeek502:
Related to https://github.com/luvit/lit/issues/200#issuecomment-276886799

This is a very quick and dirty first attempt at adding a test runner for the `coro-` lit deps. I used the `tap.lua` from [luver](https://github.com/squeek502/luver/blob/master/tests/libs/tap.lua) and included a `tests/main.lua` that will load a script and run it. The implementation is more for discussion than anything, as I'm extremely unconfident in what I was able to come up with.

The tests have to be run in a slightly weird way in order to get all the paths to match up. `luvi` has to be in the `lit` root directory, and then you can run the following:

```
luvi tests -- tests/run.lua
```

I'm also not really sure how to deal with unclosed handles, so I just manually closed the saved connection in my test case, but I'm not sure that if that's what should be done.


creationix:
This is great, thanks for starting this.

Lit has some tests currently, but they only work for me since they depend on my private github ssh keys.

squeek502:
Thinking of merging this to try to get the ball rolling. Still open to better ideas for how to implement a test runner if anyone has them, though.

Issue #210, opened 2017-06-03
WebSocket decoder bug


SinisterRectus:
Got this error using coro-wrapper and websocket-codec 3.0:

```
websocket-codec.lua:72: bad argument #1 to 'lshift' (number expected, got nil)
```
In coro-wrapper:
```lua
local item, newIndex = decode(buffer, index)
```

In websocket-codec:
```lua
local function decode(chunk, index)
  local start = index - 1
  local length = #chunk - start
  if length < 2 then return end
  local second = byte(chunk, start + 2)
  local len = band(second, 0x7f)
  local offset
  if len == 126 then
    if #chunk < 4 then return end
    len = bor(
      lshift(byte(chunk, start + 3), 8), -- line 72
      byte(chunk, start + 4))
    offset = 4
...
```
It seems that `byte(chunk, start + 3)` is rarely `nil`.

Will update with more information when I can.

SinisterRectus:
Seems to be a result of a split chunk, where only a part of one chunk is found at the end of the current buffer. I'm not sure how this happens or why the libraries fail to account for it.

SinisterRectus:
Some debug info while reading a bad buffer:

| #chunk | index | start | length | second | len |
|--------|-------|-------|--------|--------|-----|
| 4096   | 1     | 0     | 4096   | 126    | 126 |
| 4096   | 228   | 227   | 3869   | 126    | 126 |
| 4096   | 435   | 434   | 3662   | 126    | 126 |
| 4096   | 627   | 626   | 3470   | 126    | 126 |
| 4096   | 799   | 798   | 3298   | 126    | 126 |
| 4096   | 1010  | 1009  | 3087   | 126    | 126 |
| 4096   | 1201  | 1200  | 2896   | 126    | 126 |
| 4096   | 1458  | 1457  | 2639   | 126    | 126 |
| 4096   | 1629  | 1628  | 2468   | 126    | 126 |
| 4096   | 1820  | 1819  | 2277   | 126    | 126 |
| 4096   | 2077  | 2076  | 2020   | 126    | 126 |
| 4096   | 2268  | 2267  | 1829   | 126    | 126 |
| 4096   | 2479  | 2478  | 1618   | 126    | 126 |
| 4096   | 2677  | 2676  | 1420   | 126    | 126 |
| 4096   | 2848  | 2847  | 1249   | 126    | 126 |
| 4096   | 2984  | 2983  | 1113   | 126    | 126 |
[...]

PR #211, opened 2017-06-04
Fixes WebSocket decoder chunk length check

SinisterRectus:
Discovered in #210. Version bump at your discretion.


creationix:
Good catch!

Issue #212, opened 2017-06-10
HTTPS request fails for some websites (with coro-http)


edubart:
I'm using lit's coro-http for doing HTTPS requests, some websites works fine, other ones fails with the error:

`Peer aborted the SSL handshake`,

I sniffed the handshake TLS packet with wireshark and compared with other library doing HTTPS requests, I found out that the only difference was the hostname option in the TLS handshake. The solution was to add `ssl:set('hostname', host)` in `secure-socket/biowrap.lua`

```lua
function bioWrap(ctx, isServer, socket, handshakeComplete, host)
  local ssl = ctx:ssl(bin, bout, isServer)
  ssl:set('hostname', host)
```

where *host* is the host we are connecting to.

edubart:
Related to https://github.com/luvit/luvit/issues/986

PR #213, opened 2017-06-10
Adds missing hostname to ssl handle

SinisterRectus:
luvit/lit#212
luvit/luvit#986
luvit/luvit#987

### DIFF
--- a/deps/coro-net.lua
+++ b/deps/coro-net.lua
@@ -66,6 +66,8 @@ local function normalize(options, server)
   end
   if options.tls == true then
     options.tls = {}
+  end
+  if options.tls then
     if server then
       options.tls.server = true
       assert(options.tls.cert, "TLS servers require a certificate")
--- a/deps/coro-net.lua
+++ b/deps/coro-net.lua
@@ -55,14 +55,6 @@ local function normalize(options, server)
   elseif t ~= "table" then
     assert("Net options must be table, string, or number")
   end
-  if options.tls == true then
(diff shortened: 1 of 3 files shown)


SinisterRectus:
It was originally `server`, but it makes more sense as `isServer`. I made the change. As for setting the hostname, I'm assuming that it's been done right. Nothing broke on my end, so that's a plus.

SinisterRectus:
I implemented this fix thinking that I was not touching any part of the public API; however, it seems that this PR would make secure-socket incompatible with older versions of coro-net. In other words, you would need to update to the latest coro-net to use the latest secure-socket. We'll have to carefully select version numbers this time, or I can adjust the fix to something non-breaking; something more like that which is proposed in #212.

creationix:
We should probably just bump the major version on coro-net and secure-socket to be safe.

Looks good!

creationix:
I'm also fine with the suggestion in #212, it's a less invasive change and should work as well right?

SinisterRectus:
As discussed in IRC, I've restored backwards compatibility. Note that `servername` should technically be `hostname`, not `host`. Luckily, `coro-http` passes `hostname` to coro-net as `host`. This would only be a problem if anyone is using coro-net directly, and if they pass it an options table where `host` is not the same as `hostname`. We could add a fix for this with `ssl:set('hostname', servername:match("^([^:/]+)"))`, but it seems like an edge case.

PR #218, opened 2017-06-21
Fixes WebSocket chunk length check

SinisterRectus:
In #211, I replaced each `#chuck` with `length`. Apparently this line should have remained the same. The only reason why `length` can be used in the other two checks is because `length < n` is equivalent `#chunk < start + n`. Sorry for the confusion.


creationix:
Sorry I didn't catch this in review.

PR #220, opened 2017-07-20
Fix coro-spawn hanging when uv.spawn returns an error

### LINKED ISSUE #219: coro-spawn does not handle errors from uv.spawn
[This line](https://github.com/luvit/lit/blob/master/deps/coro-spawn.lua#L55):

`local handle, pid = uv.spawn(path, options, onExit)`

can error and will return `nil, errmsg` (e.g. `ENOENT: no such file or directory`). Currently, `coro-spawn` will just continue with that and the coroutine will never resume. Would submit a PR fix, but not sure how the fix should be implemented.

squeek502:
- coro-spawn will now return nil and the error message when uv.spawn errors
 - exec will also return nil and the error message when coro-spawn errors
 - fixes #219

### DIFF
--- a/deps/coro-spawn.lua
+++ b/deps/coro-spawn.lua
@@ -54,6 +54,10 @@ return function (path, options)
 
   local handle, pid = uv.spawn(path, options, onExit)
 
+  if not handle then
+    return nil, pid
+  end
+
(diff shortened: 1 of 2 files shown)


rphillips:
lgtm

Issue #224, opened 2017-09-09
coro-http.request hangs on closed connection


SinisterRectus:
On rare occasions, requests infinitely hang. It's somewhat difficult to debug the coro libs because a lot of functionality is hidden in closures, but I think I've narrowed down the issue. I'm documenting what I have here until I or someone else can discover more about this.

- Receive `connection: close` instead of `connection: keep-alive` in an HTTP response.

In `http-codec`:
- `keepAlive` becomes `false`

```lua
      elseif lowerKey == "connection" then
        head.keepAlive = lower(value) == "keep-alive"
```

- `mode` becomes `decodeRaw`
```lua
    if head.keepAlive and (not (chunkedEncoding or (contentLength and contentLength > 0)))
       or (head.method == "GET" or head.method == "HEAD") then
      mode = decodeEmpty
    elseif chunkedEncoding then
      mode = decodeChunked
    elseif contentLength then
      bytesLeft = contentLength
      mode = decodeCounted
    elseif not head.keepAlive then
      mode = decodeRaw
    end
```
```lua
  function decodeRaw(chunk, index)
    if #chunk < index then return end
    return sub(chunk, index)
  end
```

In `coro-http`:

- `read` returns `nil` for `item`
```lua
  body = {}
  if req.method == "HEAD" then
    connection.reset()
  else
    while true do
      local item = read()
      if not item then
        res.keepAlive = false
        break
      end
      if #item == 0 then
        break
      end
      body[#body + 1] = item
    end
  end
```

- `keepAlive` is `false`, so the connection is closed by calling `write` with no arguments, but it never returns.
```lua
  if res.keepAlive then
    saveConnection(connection)
  else
    write()
  end
```

SinisterRectus:
The following reproduces the issue:

```lua
local weblit = require('weblit')

weblit.app.bind {host = '127.0.0.1', port = 1337}

.use(require('weblit-auto-headers'))

.route({
    method = "DELETE",
    path = "/foo",
}, function (req, res, go)
    if req.query and req.query.q == '1' then
        res.headers['Connection'] = 'Close'
    end
    res.code = 204
    res.body = nil
end)

.start()

local http = require('coro-http')

coroutine.wrap(function()
    p('req1', http.request("DELETE", "http://127.0.0.1:1337/foo"))
    p('req2', http.request("DELETE", "http://127.0.0.1:1337/foo?q=1"))
end)()
```

The same issue occurs; `write` never returns here:

```lua
[...]

PR #227, opened 2017-10-10
Adds shutdown check to coro-channel close

SinisterRectus:
Addresses issue #224.


SinisterRectus:
Notice that previously only `err` was returned when it should have been `bool, err`. I made this change in addition to checking for a successful shutdown. I don't know why certain shutdowns fail.

PR #231, opened 2017-12-16
Add res.redirects and res.url

freshtacos:
res.redirects -> table of all redirects
res.url -> final destination url


SinisterRectus:
Sorry this PR was neglected. This change looks okay to me. Can you give an example of it working as intended? Also, this would need a minor version bump.

PR #236, opened 2018-04-29
updated so script should work

takase1121:
script can now choose tls11 tls12


squeek502:
Duplicate of #235, which was just merged. Thanks, though!

squeek502:
Actually, this is different than #235, as this adds the fix to `get-lit.ps1` directly, meaning it would fix the error when running `get-lit.ps1` directly.

PR #241, opened 2018-09-12
Cleanup handles properly on exit;

bjornbytes:
When running lit on macOS, I noticed a problem where output was being truncated:

```
$ luvi . -- --help
lit version: 3.6.0
luvi version: v2.7.6
command: --help

Lit CLI Commands
================
 lit help                    Show usage information.
 lit version                 Show just the lit version.
 lit update                  Update lit to the latest version.
 lit make appdir             Build appdir into a single executable.
 lit serve prefix            Start a lit package server (upstream or proxy).
                             `prefix` is the protocol and host.
                             EG: lit serve https://lit.luvit.io
Local configuration
-------------------
 lit auth username           Verify local private key and set username.
 lit up [url]                Go online.  Uses default upstream or url.
 lit down                    Go offline (disable upstream).
 lit config                  Print configuration.

Package Management
------------------
 lit init                    Bootstrap a new package.lua file or an init.lua
                             entry file
 lit ls [path]               List packages in folder or zip.
 lit install%$
```

I didn't notice this behavior with the main luvit binary, and after some investigation found that it had some logic to close handles on exit so they get properly flushed.  Copying this over to lit fixed the output truncation.

### DIFF
--- a/main.lua
+++ b/main.lua
@@ -26,6 +26,23 @@ local aliases = {
   ["-h"] = "help",
 }
 
+local function exit(status)
+  uv.walk(function(handle)
+    if handle then
+      local function close()
+        if not handle:is_closing() then handle:close() end
+      end
+      if handle.shutdown then
+        handle:shutdown(close)
+      else
+        close()
+      end
+    end
+  end)
+  uv.run()
+  os.exit(status)
+end
+
 _G.p = require('pretty-print').prettyPrint
 local version = require('./package').version
 coroutine.wrap(function ()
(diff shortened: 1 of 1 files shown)


bjornbytes:
Another thought I had about this: Since this handle-closing behavior ended up being necessary for 2 of the big luv(i)-based apps, it seems like it could be a "sensible default" type of thing.  Maybe luvi could provide this out of the box, expose a helper function for it, or at least document it somewhere.

PR #245, opened 2019-02-15
Update core.lua

tbmale:
Hi,

  The latest core generate an error when make-ing
`[string "bundle:libs/core.lua"]:383: attempt to index field 'luvi' (a nil value)`

This will just walk around the error (until someone smarter will truly find it).


squeek502:
To add a bit more detail:

It seems like this error will happen when the target package does not have ` luvi` key in its `package.lua`/lit metadata. See the [`luvi` key in `lit`'s `package.lua`](https://github.com/luvit/lit/blob/master/package.lua#L9-L12).

This PR seems like a good enough fix for that error.

PR #254, opened 2019-03-23
Make server calc missing snapshot deps more consistently

### LINKED ISSUE #253: Lit v3.5.4 zip failing
Getting the zip of Lit v3.5.4 from lit.luvit.io is failing the assertion added in #250:

https://lit.luvit.io/packages/luvit/lit/v3.5.4.zip

> "[string \"bundle:libs/api.lua\"]:293: Snapshot missing and resolved snapshot hash differs from existing hash (hash=fde2b65eb81079aab53f6bb62c72e6ead6ef8423, resolved=62380aefa2a6c10e79f156e11dd9908af13209ca)"

This in turn causes the Luvit CI to fail: https://travis-ci.org/luvit/luvit/jobs/508910964

squeek502:
When creating the zip of a missing snapshot, makes the server calc deps by only including deps that are older than the package that is including them. This should usually result in the same hash as the exising snapshot hash.

Fixes #253

EDIT: Note that this currently only works properly if the local db has tags for all of the versions of every dep that should be used--e.g. if you only have newer tags locally then this will fail (see fetch.lua command below for how I was fixing this for local testing). It would be best if `calculate-historic-deps` grabbed any missing versions from the remote before calculating the deps, but for now that's not super necessary (since this is mostly a hotfix). There's a TODO comment in the code for where this could be done.

### DIFF
--- a/libs/api.lua
+++ b/libs/api.lua
@@ -77,7 +77,7 @@ local jsonStringify = require('json').stringify
 local jsonParse = require('json').parse
 local modes = require('git').modes
 local exportZip = require('export-zip')
-local calculateDeps = require('calculate-deps')
+local calculateHistoricDeps = require('calculate-historic-deps')
 local queryDb = require('pkg').queryDb
 local installDeps = require('install-deps').toDb
 local ffi = require('ffi')
@@ -284,7 +284,7 @@ return function (db, prefix)
 
       if not snapshotExists then
         local deps = {}
-        calculateDeps(db, deps, meta.dependencies)
+        calculateHistoricDeps(db, deps, meta.dependencies)
         hash = installDeps(db, hash, deps)
       end
 
--- a/libs/calculate-historic-deps.lua
+++ b/libs/calculate-historic-deps.lua
@@ -1,0 +1,118 @@
+--[[
+
+Copyright 2014-2015 The Luvit Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS-IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+--]]
+
+local normalize = require('semver').normalize
+local gte = require('semver').gte
+local log = require('log').log
+local queryDb = require('pkg').queryDb
+local colorize = require('pretty-print').colorize
+
+return function (db, deps, newDeps, date)
+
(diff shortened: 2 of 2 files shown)


squeek502:
For testing purposes, I was using the following commands (put in `lit/commands`):

snapshot.lua
Usage: `lit snapshot <author> <name> <version>` (i.e. `lit snapshot luvit lit v3.5.4`)
Checks if the snapshot exists. If it doesn't tries to generate it using the same dep resolution as this PR
```lua
return function()
  local config = require('autoconfig')()
  local makeDb = require('db')
  local pkg = require('pkg')
  local log = require("log").log
  local calculateDeps = require('calculate-deps')
  local calculateHistoricDeps = require('calculate-historic-deps')
  local installDeps = require('install-deps').toDb

  local core = require('core')()
  local db = core.db
  local author_arg = assert(args[2], "missing author argument")
  local name_arg = assert(args[3], "missing name argument")
  local version_arg = assert(args[4], "missing version argument")

  print("")
  if version_arg then
    local author = author_arg
    local name = name_arg
    local version = version_arg
    local fullname = author .. "/" .. name .. "@" .. version
    local tagHash = db.read(author, name, version)
    if not tagHash then
      log("fetching tag from remote", fullname)
      local hash = db.readRemote(author, name, version)
      assert(hash, fullname.." not found in remote lit db")
      assert(db.fetch({hash}))
      tagHash = hash
    end
    assert(tagHash, fullname .. " not found")
    local meta, kind, treeHash = pkg.queryDb(db, tagHash)
    local tag = db.loadAs('tag', tagHash)
    log("tag date", tag.tagger.date.seconds)
    local snapshotHash = meta.snapshot
    log("snapshot hash", snapshotHash)

    local snapshotExists = false
    if meta.snapshot then
      snapshotExists = pcall(function() db.loadAny(meta.snapshot) end)
    end

    local resolvedHash
    if not snapshotExists then
      log("snapshot missing", "resolving deps")
      local deps = {}
      calculateHistoricDeps(db, deps, meta.dependencies, tag.tagger.date.seconds)
      resolvedHash = installDeps(db, treeHash, deps)
      log("resolved hash", resolvedHash)
    end

    -- Make sure the resolved snapshot hash matches
    if not snapshotExists then
      assert(resolvedHash == snapshotHash, "Snapshot missing and resolved snapshot hash differs from existing hash (hash="..snapshotHash..", resolved="..resolvedHash..")")
    end

    assert(pkg.queryDb(db, snapshotHash))
  end
  print("")
end
```

fetch.lua
Usage: `lit fetch <author> <name> <version>` (i.e. `lit fetch luvit querystring v2.0.0`)
Fetches a version from the remote db and adds it to the local db
```lua
return function()
  local core = require('core')()
  local db = core.db
  local author = assert(args[2], "missing author argument")
  local name = assert(args[3], "missing name argument")
  local version = assert(args[4], "missing version argument")

  print("")
  local hash = db.readRemote(author, name, version)
  assert(hash, "version not found")
  db.fetch({hash})
  print("")
end
```

PR #260, opened 2019-06-09
add a special case to coro-fs.readFile to allow it to work properly o…

### LINKED ISSUE #259: coro-fs readFile fails on /proc files
When using coro-fs on files in /proc, readFile incorrectly returns empty strings.

Using the luvit fs readFileSync and readFile functions works correctly and returns the correct contents, matching what is produced by `cat`.

Some files this bug can be reproduced on are /proc/stat, /proc/cgroups, /proc/modules, /proc/mounts

aiverson:
…n virtual files on linux.

Resolves #259

### DIFF
--- a/deps/coro-fs.lua
+++ b/deps/coro-fs.lua
@@ -1,6 +1,6 @@
 --[[lit-meta
   name = "creationix/coro-fs"
-  version = "2.2.2"
+  version = "2.2.3"
   homepage = "https://github.com/luvit/lit/blob/master/deps/coro-fs.lua"
   description = "A coro style interface to the filesystem."
   tags = {"coro", "fs"}
@@ -9,6 +9,7 @@
     "creationix/pathjoin@2.0.0"
   }
   author = { name = "Tim Caswell" }
+  contributors = {"Tim Caswell", "Alex Iverson"}
 ]]
 
 local uv = require('uv')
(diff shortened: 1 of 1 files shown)


deps/coro-fs.lua
```diff
@@ -130,7 +131,22 @@ function fs.readFile(path)
...
+      -- handle magic files the kernel generates as requested.
+      -- hopefully the heuristic works everywhere
+      local buffs = {}
+      local offs = 0
+      local size = 1024 * 48
+      repeat
+        data, err = fs.read(fd, size, offs)
+        table.insert(buffs, data or "")
+      until err or #data < size
+      return table.concat(buffs), err
```
squeek502: This should return `nil, err` on error rather than `"", err`. Also this early return means that `uv.fs_close` is never called.

Should probably change this line to something like
```lua
if not err then
  data = table.concat(buffs)
end
```
instead of returning early.

deps/coro-fs.lua
```diff
@@ -130,7 +131,22 @@ function fs.readFile(path)
...
   stat, err = fs.fstat(fd)
   if stat then
-    data, err = fs.read(fd, stat.size)
+    --special case files on virtual filesystem
+    if stat.size == 0 and stat.birthtime.sec == 0 and stat.birthtime.nsec == 0 then
+      -- handle magic files the kernel generates as requested.
+      -- hopefully the heuristic works everywhere
+      local buffs = {}
+      local offs = 0
+      local size = 1024 * 48
```
squeek502: Any reason why this chunk size was chosen? Luvit uses 8192.

deps/coro-fs.lua
```diff
@@ -130,7 +131,22 @@ function fs.readFile(path)
...
+      repeat
+        data, err = fs.read(fd, size, offs)
```
squeek502: Won't this read the same chunk over and over? `offs` never changes.

aiverson: Yes, good point. I didn't test on a virtual file larger than 48 kilobytes. Fixing.

deps/coro-fs.lua
aiverson: Done.

deps/coro-fs.lua
aiverson: Fixed

PR #267, opened 2020-02-28
api: Do more caching to speed up search responses

squeek502:
Should help with https://github.com/luvit/luvit.io/issues/24

There are two parts to this:

- The API was already caching the `hash -> meta` part of the `author/name/version -> hash -> meta` lookup, but this cache was not being warmed on start, so the first request would take roughly twice as long as future requests.
  + Warming the cache in advance also required a change to how urls were cached, since its technically possible for `prefix` to change between requests, so we can't use the same cached url every time
- All db lookups would access the file system every time, meaning that all `/search/` endpoint requests would iterate all refs, read them, etc for general searches i.e. `/search/*`. The actual values very rarely change, though, and the amount of memory needed to cache all those lookups is rather small (~1.5MB for lit-backup.git), so we can cache them and just invalidate the cache when a new ref is added via `db.write`.
  + This cache is also warmed at the same time as the `hash -> meta` one

Performance numbers for `/search/*`:

Before:
```
request 1 handled in 402ms, body length = 182518
request 2 handled in 209ms, body length = 182518
request 3 handled in 216ms, body length = 182518
request 4 handled in 206ms, body length = 182518
request 5 handled in 205ms, body length = 182518
```

After:
```
cache warmed in 444ms, using 1.52 MB memory
request 1 handled in 22ms, body length = 182518
request 2 handled in 21ms, body length = 182518
request 3 handled in 18ms, body length = 182518
request 4 handled in 19ms, body length = 182518
request 5 handled in 18ms, body length = 182518
```

(almost all the time for `/search/*` is now spent in `json.stringify`)

---

This is currently a draft because I haven't tested the cache invalidation part of it yet (`db.write`), and am not positive that `db.write` is the only pathway to modifying the cached data (if other `db` functions could modify the cached data, then the cache would not be updated, leading to the lit api responses being out-of-sync with the actual state of the db).

### DIFF
--- a/commands/serve.lua
+++ b/commands/serve.lua
@@ -5,9 +5,12 @@ return function ()
   local log = require('log').log
   local makeRemote = require('codec').makeRemote
   local core = require('core')()
+  local cachedDb = require('db-cached')
 
   local handlers = require('handlers')(core)
-  local handleRequest = require('api')(core.db, args[2])
+  -- use cached db
+  local db = cachedDb(core.db)
+  local handleRequest = require('api')(db, args[2])
 
   local app = require('weblit-app')
   require 'weblit-websocket'
--- a/libs/api.lua
+++ b/libs/api.lua
@@ -468,6 +468,10 @@ return function (db, prefix)
           end
         end
       end
+      -- update all urls to use the current prefix
+      for match, meta in pairs(matches) do
+        meta.url = (prefix or '') .. urlCache[meta]
+      end
       local res = {
         query = query,
         matches = matches,
@@ -523,10 +527,6 @@ return function (db, prefix)
       end
     end
     if type(body) == "table" then
-      -- update all urls to use the current prefix
-      for match, meta in pairs(body.matches) do
-        meta.url = (prefix or '') .. urlCache[meta]
-      end
       body = jsonStringify(body) .. "\n"
       res.headers["Content-Type"] = "application/json"
     end
--- a/libs/api.lua
+++ b/libs/api.lua
@@ -83,6 +83,8 @@ local installDeps = require('install-deps').toDb
 local ffi = require('ffi')
 local fs = require('coro-fs')
 local metrics = require('metrics')
+local cachedDb = require('db-cached')
+local uv = require('uv')
 
 local function hex_to_char(x)
   return string.char(tonumber(x, 16))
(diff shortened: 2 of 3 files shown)


squeek502:
Code used for benchmarking (put in `bench.lua` in the `commands` dir and run via `luvi . -- bench`):

```lua
local args = {...}
return function ()
  local uv = require 'uv'
  local core = require('core')()
  local db = core.db
  local handleRequest = require('api')(db, '')

  local req = {
    method = "GET",
    path = "/search/*",
    headers = {Host = "localhost"},
  }

  local lastBody = nil
  for i=1,5 do
    local res = {
      headers = {},
    }

    local start = uv.now(uv.update_time())
    handleRequest(req, res)
    local time = uv.now(uv.update_time()) - start
    print('request '..i..' handled in ' .. time .. 'ms, body length = ' .. #res.body)
    assert(not lastBody or res.body == lastBody, 'response body mismatch between req ' .. (i-1) .. ' and req ' .. i)
    lastBody = res.body
  end
end
```

squeek502:
Okay, I confirmed that cache invalidation works when a package is published. Tested via:

- copied `litdb.git` to `litdb-server.git`
- made a separate `litconfig-server` with the `database` key set to the path to `litdb-server.git`
- set env var `LIT_CONFIG` to the path to `litconfig-server` and did `luvi . -- serve` (so the server uses the `litdb-server.git` repo)
- unset the `LIT_CONFIG` var
- `luvi . -- up ws://localhost:4822` to set the upstream in the default litconfig to the local one we just setup via `serve`
- Add/publish a new package via `luvi . -- publish test.lua` to the local upstream

With some debug prints added in `db.write` in `db-cached.lua`, the server outputs:

```
client command: 127.0.0.1:40849 - send
'db.write'      'squeek502'     'test'  '0.0.6' '27c68bcf34ce2619f3469c55f7ce41c1e712d46e'
'db.write versions cache before:'       { '0.0.1', '0.0.2', '0.0.3', '0.0.4', '0.0.5' }
'db.write versions cache after:'        { '0.0.1', '0.0.2', '0.0.3', '0.0.4', '0.0.5', '0.0.6' }
new package: squeek502/test/v0.0.6
client disconnected: 127.0.0.1:40849
```

and the new package shows up everywhere you'd expect (`/search/name:test`, `/packages/squeek502`, etc).

This is ready for review.

Issue #216, opened 2017-06-15
coro-http has no timeout


edubart:
Coro http has no timeout option, if you do a HTTP request to an unresponsive address the coroutine it will never resume.

Test case:
```lua
local http = require 'coro-http'
coroutine.wrap(function()
  http.request('GET', 'http://123.123.123.123/')
end)()
```

Running this example with luvit it will run forever.

This is problematic because I was running a daemon using coro-http and it got stuck forever when my network was down, even after the network became alive again.

halfnelson:
I hit the same problem and I made some changes to my copy of the the core coro-wrapper and coro-channel libraries to provide support for optional read timeouts. They would need to be wired through to http library, but shouldn't be too difficult. Just look for :read in there.

creationix:
@halfnelson would you like to send a PR to see if your changes can get merged here?

PR #270, opened 2020-03-26
Handle timeout requests

Lautenschlager-id:
When you set a timeout value in `coro-http`.`request`, it can happen that the returned value for `write` is a string `"timeout"`, and it ends up breaking in `write(req)`.

Now, it returns a `nil` header with body `"timeout"`.

Refers to
```
Uncaught Error: .../deps/coro-http.lua:137: attempt to call local 'write' (a string value)
```
May be helpful for https://github.com/luvit/lit/issues/216

### DIFF
--- a/deps/coro-http.lua
+++ b/deps/coro-http.lua
@@ -105,8 +105,11 @@ end
 local function request(method, url, headers, body, timeout)
   local uri = parseUrl(url)
   local connection = getConnection(uri.hostname, uri.port, uri.tls, timeout)
-  local read = connection.read
   local write = connection.write
+  if write == "timeout" then
+    return nil, write
+  end
+  local read = connection.read
 
   local req = {
     method = method,


SinisterRectus:
#216 should have been closed with #233.

Does this always happen? Can you provide code to reproduce the issue?

Lautenschlager-id:
Yes, I saw your PR but unfortunately it does not handle connection.write when it is string.

You can reproduce it with the code provided in #216 
```Lua
local http = require("coro-http")
coroutine.wrap(function()
  http.request("GET", "http://123.123.123.123/", nil, nil, 1000) -- Sets the timeout to 1s
end)()
```

SinisterRectus:
I don't get your error. I get:
```
C:\Users\Sinister\Desktop>luvit test.lua
Uncaught Error: C:/Users/Sinister/Desktop/deps/coro-http.lua:74: timeout
stack traceback:
        [builtin#2]: at 0x7ff7df1e9eb0
        [C]: in function 'run'
        [string "bundle:/init.lua"]:52: in function <[string "bundle:/init.lua"]:47>
        [C]: in function 'xpcall'
        [string "bundle:/init.lua"]:47: in function 'fn'
        [string "bundle:deps/require.lua"]:310: in function <[string "bundle:deps/require.lua"]:266>
```

Lautenschlager-id:
Oh, I see `getConnection` uses `assert` now :thinking: 

When did that happen? I have been using a pretty recent version and git blame says it is there for 4 yearslol

SinisterRectus:
Not sure. I do know that there is a discrepancy between coro-http (which uses assert) and coro-websocket (which returns nil, err).

PR #257, opened 2019-03-31
Fix secure-socket for OpenSSL > 1.1.0

squeek502:
Fixes #256

Lit version of https://github.com/luvit/luvit/pull/1062

Tested and confirmed to work with both OpenSSL 1.0.2 and 1.1.1

PR #272, opened 2020-05-14
Fix secure-socket for OpenSSL 1.1.0

truemedian:
The fix in #257 makes DEFAULT_SECUREPROTOCOL `TLS` on OpenSSL 1.1.0, which does not support it (it correctly does this for OpenSSL < 1.1.0 and OpenSSL > 1.1.0). This specific OpenSSL version does not accept `TLS` and should instead use `SSLv23` like previous versions.

This issue keeps coming up with people using Luvi 2.9.1 (the affected luvi).

### DIFF
--- a/deps/secure-socket/context.lua
+++ b/deps/secure-socket/context.lua
@@ -33,7 +33,7 @@ do
   local isLibreSSL = V:find('^LibreSSL')
 
   _, _, V = openssl.version(true)
-  local isTLSv1_3 = not isLibreSSL and V > 0x10100000
+  local isTLSv1_3 = not isLibreSSL and V > 0x10101000
 
   if isTLSv1_3 then
     DEFAULT_SECUREPROTOCOL = 'TLS'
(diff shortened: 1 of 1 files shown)


deps/secure-socket/context.lua
```diff
@@ -33,7 +33,7 @@ do
   local isLibreSSL = V:find('^LibreSSL')
 
   _, _, V = openssl.version(true)
-  local isTLSv1_3 = not isLibreSSL and V > 0x10100000
+  local isTLSv1_3 = not isLibreSSL and V > 0x10101000
```
zhaozg, line 36: Thats means need openssl 1.1.1, not 1.1.0

squeek502: That seems right, see the [Major changes between OpenSSL 1.1.0i and OpenSSL 1.1.1 section here](https://www.openssl.org/news/openssl-1.1.1-notes.html)

> - Support for TLSv1.3 added (see https://wiki.openssl.org/index.php/TLS1.3 for further important information).

squeek502, line 36:
```suggestion
  local isTLSv1_3 = not isLibreSSL and V >= 0x10101000
```

squeek502:
Thanks for the fix!

PR #276, opened 2020-10-01
Preserves tracebacks  for coroutine resumes

### LINKED ISSUE #273: coro libraries naively assert coroutine resumes, original stack trace is lost.
A simple example of this would be a simple websocket request, where the error pops up in coro-channel:
```lua
local ws = require 'coro-websocket'

local function throw()
    error('this')
end

coroutine.wrap(function()
    local options = ws.parseUrl("ws://echo.websocket.org/")
    local res, read, write = ws.connect(options)
    assert(res)

    write { opcode = 1, payload = 'hello, world' }

    for chunk in read do
        throw()
    end
end)()
```
Which gives a very confusing error, you can see where the original error came from: in this case: line 4, but not the function or stack trace to get to that error.
```
Uncaught Error: .../deps/coro-channel.lua:62: .../init.lua:4: this
stack traceback:
        [C]: in function 'assert'
        .../deps/coro-channel.lua:62: in function <.../deps/coro-channel.lua:54>
        [C]: in function 'run'
        [string "bundle:/init.lua"]:52: in function <[string "bundle:/init.lua"]:47>
        [C]: in function 'xpcall'
        [string "bundle:/init.lua"]:47: in function 'fn'
        [string "bundle:deps/require.lua"]:310: in function <[string "bundle:deps/require.lua"]:266>
```
This happens because in coro-channel, the coroutine.resume is asserted, which will ignore and throw away the stack used by the coroutine in favor of the stack used by the resumer (in this case, the main thread).

SinisterRectus:
Closes #273. 

I replaced every `assert(coroutine.resume(thread, ...))` with an `assertResume(thread, ...)` wrapper defined in each file that needs it. Not really a sustainable solution, but I didn't want to make a new dependency at this time.

### DIFF
--- a/deps/coro-channel.lua
+++ b/deps/coro-channel.lua
@@ -10,6 +10,13 @@
 
 -- local p = require('pretty-print').prettyPrint
 
+local function assertResume(thread, ...)
+  local success, err = coroutine.resume(thread, ...)
+  if not success then
+    error(debug.traceback(thread, err), 0)
+  end
+end
+
 local function makeCloser(socket)
   local closer = {
     read = false,
@@ -59,7 +66,7 @@ local function makeRead(socket, closer)
       local thread = queue[dindex]
       queue[dindex] = nil
       dindex = dindex + 1
-      assert(coroutine.resume(thread, unpack(data)))
+      assertResume(thread, unpack(data))
     else
       queue[dindex] = data
       dindex = dindex + 1
@@ -116,7 +123,7 @@ local function makeWrite(socket, closer)
   local function wait()
     local thread = coroutine.running()
     return function (err)
-      assert(coroutine.resume(thread, err))
+      assertResume(thread, err)
     end
   end
 
--- a/deps/coro-channel.lua
+++ b/deps/coro-channel.lua
@@ -1,6 +1,6 @@
 --[[lit-meta
   name = "creationix/coro-channel"
-  version = "3.0.1"
+  version = "3.0.2"
   homepage = "https://github.com/luvit/lit/blob/master/deps/coro-channel.lua"
   description = "An adapter for wrapping uv streams as coro-streams."
   tags = {"coro", "adapter"}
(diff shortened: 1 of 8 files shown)


squeek502 (approved):

SinisterRectus:
I should point out that this `assertResume` doesn't return any values given by`coroutine.resume`. This isn't an actual issue, though, because all of the `coroutine.yield` calls are "empty".

truemedian:
If this were to end up going into a library instead of being copied around everywhere, I think `luvit/utils` would be the best place for it. It already holds the utils used around luvit to bind functions, I think this might fit there.

SinisterRectus:
I was actually going to do that next, for the couple of places that luvit resumes coroutines. There aren't many, but one in particular is `timer.sleep`. The `utils` module is definitely amenable to this kind of change, but does it make sense to depend on it in lit and the coro libraries?

SinisterRectus:
A general solution probably should have a mechanism for collecting the returns, although I'm not sure how useful this is when all of the resumers are callbacks that return into the void.

Example:
```lua
local pack = table.pack or function(...)
  return {..., n = select('#', ...)}
end

local function assertResume(thread, ...)
  local success, err = coroutine.resume(thread, ...)
  if not success then
    error(debug.traceback(thread, err), 0)
  end
end

local function assertResumePack(thread, ...)
  local ret = pack(coroutine.resume(thread, ...))
  if not ret[1] then
    error(debug.traceback(thread, ret[2]), 0)
  else
    return unpack(ret, 2, ret.n)
  end
end

local thread1 = coroutine.create(function(...)
  return coroutine.yield(...)
end)

local thread2 = coroutine.create(function(...)
  return coroutine.yield(...)
end)

local x = assertResume(thread1, 1)
assert(x == nil)

local a, b, c = assertResumePack(thread2, 1, nil, 3)
assert(a == 1)
assert(b == nil)
assert(c == 3)

```

SinisterRectus:
Forgot to bump the versions. Did a patch bump for each.

PR #277, opened 2020-10-18
weblit app prints out bindings twice

PatrickDahlin:
Removed extra printout of binding since it unnecessarily prints bindings twice

### DIFF
--- a/deps/weblit-server.lua
+++ b/deps/weblit-server.lua
@@ -158,13 +158,6 @@ local function newServer(run)
 
     print("Weblit server listening at:")
 
-    for i = 1, #bindings do
-      local options = bindings[i]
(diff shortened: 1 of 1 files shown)


creationix:
If I remember, these aren't always the same thing and so the code is being lazy and always printing both forms in case they don't match.  We could probably store all the results in a set (store the strings as table keys with dummy values) and filter out duplicates before printing.

truemedian:
As Tim mentioned above, this behavior is mostly intentional. Perhaps they could be filtered but dropping them altogether may not be desirable.

Bilal2453:
I've looked into filtering them out once, I can't remember the details but I figured that was unwanted as well. It is fine as it is.

PR #278, opened 2020-10-18
Add options table to coro-http request

JohnnyMorganz:
*Continuation of https://github.com/luvit/lit/pull/230*

coro-http currently does not have the functionality to prevent following redirects.

This PR adds an `options` argument to the request function, which is a table of possible options for the request
This also replaces the `timeout` argument, which has now been moved into the `options` table, making it look like:
```
options
    timeout (number) - Default: nil
    followRedirects (boolean) - Default: true
```
This also allows more options to be added over time as required.

For backwards compatibility, if a `number` is provided instead of a table to the options parameter, it is assumed as the timeout.

### DIFF
--- a/deps/coro-http.lua
+++ b/deps/coro-http.lua
@@ -1,6 +1,6 @@
 --[[lit-meta
   name = "creationix/coro-http"
-  version = "3.1.0"
+  version = "3.2.0"
   dependencies = {
     "creationix/coro-net@3.0.0",
     "luvit/http-codec@3.0.0"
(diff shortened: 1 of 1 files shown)


squeek502 (approved):
Looks good, thanks for doing this.

JohnnyMorganz:
Noticed a bug in how I handled if customOptions was a number, this should be fixed

SinisterRectus:
Looks good to me. Can you bump package.version to 3.2.0?

JohnnyMorganz:
Apologies for the delay again, bumped package.version

PR #281, opened 2021-02-12
correcting fs.access parameter name

Bilal2453:
image
`flags` is still valid, though it isn't that accurate.

Thanks for your time, and sorry for the useless PRs 😬


creationix (approved):

creationix:
This is not useless, thanks!

Issue #274, opened 2020-08-04
coro-net/coro-http server leaks connections?


u3shit:
I have a small app that uses `coro-http` directly to create a small http server. After some time I noticed in `netstat` that there are a large number of connections stuck in `CLOSE_WAIT` state, indicating that the server never actually closed the connection.
I'm not exactly sure whose responsibility is to close the server socket, but based on how `coro-http` uses the callback, I can't really close it from there: https://github.com/luvit/lit/blob/55f47c6c657127d80fd8db19e86990bde72b7de4/deps/coro-http.lua#L35-L39
So I think either `coro-net` or `coro-http` should close the socket, probably `coro-net` somewhere around here https://github.com/luvit/lit/blob/55f47c6c657127d80fd8db19e86990bde72b7de4/deps/coro-net.lua#L169

PatrickDahlin:
Looking closer at it there should be some timeout for disconnecting, if not default then at least as Keep-Alive header defines.
I can also confirm there are connections left in the CLOSE_WAIT state on my machine.

Bilal2453:
Looking back at this, nope, there are no closing of any kind explicitly happening in the code. Even if the peer aborts the connection the handle should still be bound as far as I can tell. The `break` statement you see is only to stop reading from the socket, it doesn't necessarily mean the handle is closed.
Since coro-http expose the socket handle that should help a bit until a proper closing is done.

This most likely should be implemented on coro-net side, since I can't really see a reason that would let you keep a dead connection bound, though it is as well easily implemented on coro-http side just before that `break`.

creationix:
If it's not closing the sockets, then it's a bug.  I thought I had logic to auto-close in coro-net, but maybe that broke during one of the past refactors.

Bilal2453:
`net.connect` does auto-close connection properly. `net.createServer` doesn't as far as I can tell, there is just no logic for it (or I am blind).

creationix:
I believe this is what should close it when there is an error in reader or writer or both the reader and writer have ended.

https://github.com/luvit/lit/blob/master/deps/coro-channel.lua#L47

Bilal2453:
does that get triggered on connection death? that's keepAlive state change

creationix:
It checks in all the cases that could trigger a close.

The first two are during the libuv read callback.  If there is an error or EOF, then it triggers `closer.check()`

We should get these when a client does either a graceful disconnect or a hard disconnect while we have the `onRead` callback set in libuv.

```lua
[...]

PR #283, opened 2021-03-06
coro-http: createServer: exposing the server handle

Bilal2453:
Since `coro_net.createServer` already exposes the server handle, wouldn't also make sense to do that on the `coro-http` side? So stuff such as closing the server (for example) can be done.


squeek502 (approved):

Bilal2453:
Could fix #274 as well.

PR #286, opened 2021-06-24
coro-http: properly close the socket

### LINKED ISSUE #274: coro-net/coro-http server leaks connections?

Bilal2453:
Closes #274. Forcibly close the socket stream on connection death, that's, when keepAlive is false.

On my end, this seems to solve the CLOSE_WAIT issue when `netstat`ing and the handles seems to be completely closed.
I would argue this should actually go to coro-net not to coro-http, since that's more general and affects more stuff. Though I am not sure if any other libraries do in fact expect the handle to always stay there even on CLOSE_WAIT state, otherwise that would be a breaking change.

### DIFF
--- a/deps/coro-http.lua
+++ b/deps/coro-http.lua
@@ -35,9 +35,8 @@ local function createServer(host, port, onConnect)
       head, body = onConnect(head, body, socket)
       write(head)
       if body then write(body) end
-      write("")
+      write()
       if not head.keepAlive then
-        socket:close()
         break
       end
     end
--- a/deps/coro-http.lua
+++ b/deps/coro-http.lua
@@ -1,6 +1,6 @@
 --[[lit-meta
   name = "creationix/coro-http"
-  version = "3.2.0"
+  version = "3.2.1"
   dependencies = {
     "creationix/coro-net@3.0.0",
     "luvit/http-codec@3.0.0"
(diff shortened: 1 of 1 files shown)


Bilal2453:
Since conversation in #274 lead to
> coro-channel should be in fact closing them

situation, I will be debugging why the intended behavior doesn't occur then fix that in this PR.

creationix:
This isn't quite right, but it got me very close to the real solution (I hope)

creationix:
See #289 for what I hope is the proper fix.

Bilal2453:
> This isn't quite right, but it got me very close to the real solution (I hope)

Oh ye, that's why I was waiting your confirmation on the `write("")` thing xD

> See #289 for what I hope is the proper fix.

Should do I guess, will be doing a bit of testing on that around tomorrow, Thanks!

PR #288, opened 2021-06-25
Add proper stateful codec support to coro-net.

creationix:
http-codec is a stateful parser and this wasn't really supported properly.
Before this patch, coro-net would reuse the exact same stateful parser
for all concurrent and future streams which has many issues.

This was probably the cause of many bugs including the one that makes
the production website hang whenever a request triggers http raw body mode.

This PR updates coro-net to now support both statefull and stateless codecs
and updates coro-http, coro-websocket and others to use the new feature.

Third-party libraries that use coro-net directly with http-codec will need
a similar update.

### DIFF
--- a/deps/coro-http.lua
+++ b/deps/coro-http.lua
@@ -1,8 +1,8 @@
 --[[lit-meta
   name = "creationix/coro-http"
-  version = "3.2.0"
+  version = "3.2.1"
   dependencies = {
-    "creationix/coro-net@3.0.0",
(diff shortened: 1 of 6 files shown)


creationix:
Many thanks to @Bilal2453 for helping me find the root cause!

Bilal2453:
Works like a charm! Thanks for putting time into this.

Bilal2453 (approved):

PR #290, opened 2021-06-28
coro-split: fixing coroutine resume before yielding

Bilal2453:
**Issue:**

If you try the following code
```lua
require'coro-split'(
  function()
    print(1)
    print(2)
  end,
  function()
    print(3)
    print(4)
  end,
  function()
    print(5)
    print(6)
  end
)
```
it would error with the said error message after it executed everything successfully
```
```

**Explanation:**

That's due to all tasks being spawned and completed before `coroutine.yield` have been reached, therefor the `coroutine.resume` in `check` would fail.

**Solution:**
Basically make sure you only resume if yield have been executed (or tagged as so), otherwise return without yielding.

### DIFF
--- a/deps/coro-split.lua
+++ b/deps/coro-split.lua
@@ -27,9 +27,10 @@ return function (...)
   local thread = coroutine.running()
   local left = #tasks
   local results = {}
+  local yielded = false
   local function check()
     left = left - 1
-    if left == 0 then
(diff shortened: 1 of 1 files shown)


squeek502 (approved):
Makes sense to me

creationix:
Love it, thanks!

PR #294, opened 2021-08-30
Add 'Development' section to the readme with some useful advice

squeek502:
Bit of a pet peeve of mine when information like this is left unsaid, so I thought I'd fix that.


Bilal2453:
Looks like this was mentioned in the [Luvit readme](https://github.com/luvit/luvit/blob/master/README.markdown#hacking-on-luvit-core) at the very least.

PR #296, opened 2021-09-30
secure-socket: more accurate error messages

Bilal2453:
This removes the assumption that the unverified cert is the first in the chain, if additional certs are supplied this reports an error message that looks like `Error verifying peer: ok`.

Before the changes: 
image

After the changes:
image

This might still be ambiguous to the user although, as in, the browser could still be able to access a said website -since the main cert is valid- but coro-net for example could not access it because one of the additional certs is invalid. Perhaps we can somehow report which cert is exactly unverified?

### DIFF
--- a/deps/secure-socket/biowrap.lua
+++ b/deps/secure-socket/biowrap.lua
@@ -51,7 +51,13 @@ return function (ctx, isServer, socket, handshakeComplete, servername)
       local success, result = ssl:getpeerverification()
       socket:read_stop()
       if not success and result then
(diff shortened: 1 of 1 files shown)


SinisterRectus (approved):
For reference:

> all certificate in chains verify result preverify_ok as boolean verify result error as number error code error_string as string error message error_depth as number verify depth current_cert as x509 certificate to verified

https://zhaozg.github.io/lua-openssl/modules/ssl.html#ssl_ctx:getpeerverification

PR #304, opened 2022-01-07
Hotfix for #303

### LINKED ISSUE #303: coro-channel: resuming before having a chance to yield
`coro_http.request("GET", "https://google.com", {}, "")`
would error with;
```
stack traceback:
	/mnt/data/bilal/deps/coro-channel.lua:16: in function 'assertResume'
	/mnt/data/bilal/deps/coro-channel.lua:126: in function 'callback'
	/mnt/data/bilal/deps/secure-socket/biowrap.lua:43: in function 'write'
	/mnt/data/bilal/deps/coro-channel.lua:148: in function 'write'
	/mnt/data/bilal/deps/coro-http.lua:149: in function </mnt/data/bilal/deps/coro-http.lua:106>
	[C]: in function 'xpcall'
	[string "bundle:/deps/repl.lua"]:97: in function 'evaluateLine'
	[string "bundle:/deps/repl.lua"]:189: in function <[string "bundle:/deps/repl.lua"]:187>
stack traceback:
	[C]: in function 'error'
	/mnt/data/bilal/deps/coro-channel.lua:16: in function 'assertResume'
	/mnt/data/bilal/deps/coro-channel.lua:126: in function 'callback'
	/mnt/data/bilal/deps/secure-socket/biowrap.lua:43: in function 'write'
	/mnt/data/bilal/deps/coro-channel.lua:148: in function 'write'
	/mnt/data/bilal/deps/coro-http.lua:149: in function </mnt/data/bilal/deps/coro-http.lua:106>
```
While the request is very wrong (providing a body on GET) it shouldn't have produced such error. I am afraid it is affecting wider use-cases of the library.

I will be looking deeper into this.

Bilal2453:
This should fix #303 by tracking the state of the thread.
We might want to patch and update the coro-channel package on Lit as fast as we can, I've seen so far 3-4 people complaining about this.

### DIFF
--- a/deps/coro-channel.lua
+++ b/deps/coro-channel.lua
@@ -1,6 +1,6 @@
 --[[lit-meta
   name = "creationix/coro-channel"
-  version = "3.0.3"
+  version = "3.0.4"
   homepage = "https://github.com/luvit/lit/blob/master/deps/coro-channel.lua"
   description = "An adapter for wrapping uv streams as coro-streams."
   tags = {"coro", "adapter"}
@@ -120,14 +120,21 @@ end
 
 local function makeWrite(socket, closer)
 
+  -- states to track when we should yield and resume
+  local hasYielded, hasReturned, returnedValue
   local function wait()
     local thread = coroutine.running()
     return function (err)
-      assertResume(thread, err)
+      if hasYielded then
+        assertResume(thread, err)
+      else
+        hasReturned, returnedValue = true, err
+      end
     end
   end
 
   local function write(chunk)
+    hasYielded, hasReturned, returnedValue = false, false, nil
     if closer.written then
       return nil, "already shutdown"
     end
@@ -151,8 +158,14 @@ local function makeWrite(socket, closer)
       closer.check()
       return nil, err
     end
-    err = coroutine.yield()
-    return not err, err
+
+    if hasReturned then
+      return not returnedValue, returnedValue
+    else
+      hasYielded = true
+      err = coroutine.yield()
+      return not err, err
+    end
   end
 
   return write


Bilal2453:
@squeek502 Can this be reviewed/merged quickly? It seems to be a major issue around

SinisterRectus:
I'll need to look more at how this is happening, but would it just be easier to reject a body for GET requests? Maybe that would be breaking, though.

Bilal2453:
I've explained well enough why this happen in #303, basically `socket:write()` returns too quickly on an empty chunk when wrapped with secure-socket, not sure what changed exactly and triggered this, but there has always been a chance of it happening eventually -effectively we are always assuming the thread was yielded which isn't always true-.

We could mitigate this with just checking the body length, that still wouldn't be a fix + we don't know just yet what else this is affecting. That do seem like more trouble and not worth it quite honestly.

Bilal2453:
> but would it just be easier to reject a body for GET requests
@SinisterRectus Oh didn't notice you specified GET requests with this mitigation. I did mention this in the issue, but here again, this is not only on GET requests, it happens on ANYTHING that uses the coro-channel wrapWrite. So the following will error too `coro_http.request("POST", "https://google.com", {}, "")`, this is even more trouble since this request should be valid.

Bilal2453:
I was debugging an issue with TLS servers when stress tested with this fix applied, and I noticed `write("")` can still cause problems, more specifically this time it can lead to yielding the write and never resuming it.
This is also for some reason *very* slow compared to a proper fix, at least 10 times slower.

since `hasReturned` can be assigned whatever value by the read, it can also be a falsy value.

A slightly better fix would to have something like:
```lua
local some_unique_value = {}
hasReturned = some_unique_value
-- ...
hasReturned = data
-- ...
if hasReturned == some_unique_value then
 -- we have resumed before the yield
else
 -- we should yield as write has not finished yet
end
```
also we can stop using the `hasYielded` upvalue and instead use `coroutine.status(thread)`.

PR #315, opened 2022-09-22
get-lit: update luvi to 2.14.0

Bilal2453:
This also workaround a mis-labled Luvi tag, v2.12.0 is in fact Luvi version `2.11.0-21` which causes an error with Lit's `core.getLuvi` not finding a tag for `2.12.0-21`.


Bilal2453:
we may also want to re-publish the binaries in tag v2.12.0 or perhaps rename it?

Bilal2453:
@truemedian should be good for a merge. note the two commits, a squash merge would be preferable

PR #316, opened 2022-11-09
Create update-lit.sh

Be1zebub:
Bash script that runs [get-lit.sh](https://github.com/luvit/lit/raw/master/get-lit.sh) when `lit/luvi/luvit` is not installed or not updated.  
  
  
I wrote this to create luvit docker image for [pterodactyl](https://pterodactyl.io/) with auto-updating feature.  
  
I made it dependent on [get-lit.sh](https://github.com/luvit/lit/raw/master/get-lit.sh), without worrying about the fact that "when one component needs update, it will update all" - due to the fact that the luvit is very small, the execution [get-lit.sh](https://github.com/luvit/lit/raw/master/get-lit.sh) takes no more than a few seconds on any machine.

### DIFF
--- a/update-lit.sh
+++ b/update-lit.sh
@@ -1,11 +1,15 @@
 #!/bin/bash
 
 cecho() {
-	# foreground \033[48;2;RRR;GGG;BBBm
-	# background \033[38;2;RRR;GGG;BBBm
-	# clear      \033[0m
+        # foreground \033[48;2;RRR;GGG;BBBm
+        # background \033[38;2;RRR;GGG;BBBm
+        # clear      \033[0m
 
-	echo "\033[38;2;50;210;255m$1\033[0m";
+        if [ -z ${BASH_VERSION} ]; then # bash interpreter
+                echo "\033[38;2;50;210;255m$1\033[0m";
+        else # sh interpreter
+                echo -e "\033[38;2;50;210;255m$1\033[0m";
+        fi
 }
 
 while test $# -gt 0; do
--- a/update-lit.sh
+++ b/update-lit.sh
@@ -5,17 +5,17 @@ cecho() {
 	# background \033[38;2;RRR;GGG;BBBm
 	# clear      \033[0m
 
-	echo -e "\033[38;2;50;210;255m$1\033[0m";
+	echo "\033[38;2;50;210;255m$1\033[0m";
 }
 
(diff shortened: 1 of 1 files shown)


Be1zebub:
image

Be1zebub (commented):
commit name typo, *latest

update-lit.sh
```diff
@@ -0,0 +1,155 @@
...
+        # clear      \033[0m
+
+        if [ -z ${BASH_VERSION} ]; then # bash interpreter
+                echo "\033[38;2;50;210;255m$1\033[0m";
+        else # sh interpreter
+                echo -e "\033[38;2;50;210;255m$1\033[0m";
+        fi
+}
+
+while test $# -gt 0; do
```
truemedian, line 15: If this is meant to be used as `curl % | bash`, then it should not have this interface.

update-lit.sh
```diff
@@ -0,0 +1,155 @@
+#!/bin/bash
```
truemedian, line 1: Please make this use `/bin/sh`, if it needs bash features then find an alternative to those features.
Also: use `set -eu` to prevent errors from propagating

update-lit.sh
```diff
@@ -0,0 +1,155 @@
...
+
+        if [ -z ${BASH_VERSION} ]; then # bash interpreter
+                echo "\033[38;2;50;210;255m$1\033[0m";
+        else # sh interpreter
+                echo -e "\033[38;2;50;210;255m$1\033[0m";
+        fi
+}
+
+while test $# -gt 0; do
+	case "$1" in
```
truemedian, line 16: These could all be handled by `LUVI_VERSION` and `LIT_VERSION` environment variables.

truemedian (commented):
In general this script seems very complicated for a relatively simple job. I feel that it would benefit greatly from being less complex and simply doing an update to whatever version you provide in LUVI_VERSION and LIT_VERSION (or the latest if not provided)

Bilal2453:
I am also not entirely sure if we really need such script in the repo.  While I am sure it will be useful for someone out there, internally I see no use for it, the only reason we have `get-lit` is the install script; it is the main way for anyone to get Luvit/Lit/Luvi.

So unless there is an internal usage for this script, I am not entirely sure if we want it here.

PR #319, opened 2023-02-19
coro-net: fix handle leak when connect errors

Bilal2453:
Coro-net creates a new handle (whether tcp socket or a pipe) then proceeds to try connecting it. The code handling the connection failure does not seem to free the recently opened handle leading to a leak happening on `socket:connect` failures.

For ease of testing, I simulated a network error by simply commenting out `uv.getaddrinfo` and forcing it to connect to a non-existence host / port, forcing a `ECONNREFUSED` error like so:

```lua
  options.isTcp = true
  if options.isTcp then
    local res = {{addr = '127.0.0.1', port = 8000}}
    if not res then return nil, err end
    socket = uv.new_tcp()
    socket:connect(res[1].addr, res[1].port, makeCallback(options.timeout))
    ...
```

The results of doing multiple `coro_net.connect` calls with this simulation resulted in the following:
image

With this patch it seems to behave:
image

One thing I am not entirely sure about if it is possible for `socket` to not be nil, yet manage to fail on `socket:close()`, I assume that is not a possible scenario.

### DIFF
--- a/deps/coro-net.lua
+++ b/deps/coro-net.lua
@@ -115,6 +115,7 @@ local function connect(options)
     if not secureSocket then secureSocket = require('secure-socket') end
     dsocket, err = secureSocket(socket, options.tls)
     if not dsocket then
+      if socket then socket:close() end
       return nil, err
     end
   else
--- a/deps/coro-net.lua
+++ b/deps/coro-net.lua
@@ -1,6 +1,6 @@
 --[[lit-meta
   name = "creationix/coro-net"
-  version = "3.3.0"
+  version = "3.3.1"
   dependencies = {
     "creationix/coro-channel@3.0.0",
     "creationix/coro-wrapper@3.0.0",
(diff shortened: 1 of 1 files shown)


Bilal2453:
I forgot to include that second close in the first commit, but I also just noticed we only really need `socket:close()`, no need for the if statement? should I change that?

deps/coro-net.lua
```diff
@@ -106,12 +106,16 @@ local function connect(options)
     socket:connect(options.path, makeCallback(options.timeout))
   end
   success, err = coroutine.yield()
-  if not success then return nil, err end
+  if not success then
+    if socket then socket:close() end
```
truemedian, line 110: `socket` is guaranteed to be non-nil here, the if is unnecessary.

deps/coro-net.lua
```diff
@@ -106,12 +106,16 @@ local function connect(options)
...
+    return nil, err
+  end
   local dsocket
   if options.tls then
     if not secureSocket then secureSocket = require('secure-socket') end
     dsocket, err = secureSocket(socket, options.tls)
     if not dsocket then
+      if socket then socket:close() end
```
truemedian, line 118: same here, `socket` will always exist.

truemedian (changes requested):

truemedian (approved):

squeek502:
Thanks!

PR #321, opened 2023-06-11
RFC 6455 Sec-WebSocket-Key

Ta-noshii:
Update amount of bytes used in Sec-WebSocket-Key header https://datatracker.ietf.org/doc/html/rfc6455#section-4.2.1

### DIFF
--- a/deps/websocket-codec.lua
+++ b/deps/websocket-codec.lua
@@ -183,8 +183,8 @@ end
 
 -- Make a client handshake connection
 local function handshake(options, request)
-  -- Generate 20 bytes of pseudo-random data
-  local key = concat({rand4(), rand4(), rand4(), rand4(), rand4()})
+  -- Generate 16 bytes of pseudo-random data
+  local key = concat({rand4(), rand4(), rand4(), rand4()})
   key = base64(key)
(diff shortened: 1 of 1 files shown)


Bilal2453:
You seem to have re-discovered this on your own, nice!   The community has actually noticed this sometime ago thanks to @GitSparTV although seems like he got busy and forgot to PR his fix (so did I).
if someone wants to read the details and walk-through of this you can [here on Discord](https://discord.com/channels/377572091869790210/379042219426054146/1065769682558255195) (there is even another bug reported there!).

By the way, that concatenation can simply be a string concatenation with `..`, as Spar did in that linked message.

Bilal2453:
@SinisterRectus Seems like this is affecting some Discordia voice clients, some Discord voice servers are strict about this header (such as the Brazil region) and do check its length, other regions do not.

SinisterRectus (approved):

Lort533:
Are we sure this is a correct solution? Because this indeed affects some Discord voice servers, and reverting this change indeed does fix the issue. That IETF link suggests the original 16 bytes, not 20. Am I not understanding something here?

PR #326, opened 2024-07-23
fix: secure-socket does not check for certificate domain

Bilal2453:
After secure-socket verifies a certificates and deems it to be valid, it does not check whether this certificate is actually provided by the same server that claims to be the owner of the certificate.  This results in a possible attack vector where an attacker that is intercepting a connection (through a MITM attack) to provide any valid SSL certificate (for example, one they made themselves for a domain they own) and secure-socket (and by extension, coro-net/coro-http etc) would accept that certificate.

1. A connection is intercepted by an attacker between `google.com` and a user-agent using secure-socket.
2. The attacker provides their valid certificate for the domain they own `doodle.com`.
3. secure-socket will check for the validity of the certificate (which it is valid), and start encrypting the packets following that.
4. The attacker can now decrypt all packets and read what is going on with the connection, potentially stealing the user-agent session tokens, passwords, etc.

A Common Weakness Enumeration for this would be [297](https://cwe.mitre.org/data/definitions/297.html):
> 
> ```c
>cert = SSL_get_peer_certificate(ssl);
> if (cert && (SSL_get_verify_result(ssl)==X509_V_OK)) {
>
> // do secret things
> }
>```
> Even though the "verify" step returns X509_V_OK, this step does not include checking the Common Name against the name of the host. That is, there is no guarantee that the certificate is for the desired host. The SSL connection could have been established with a malicious host that provided a valid certificate.

This has been mitigated in this PR by applying an identity check using [x509:check_host](https://zhaozg.github.io/lua-openssl/modules/x509.html#x509:check_host) and error if the domain of the certificate does not match the server name.

This was caught when I was (out of curiosity)  testing out https://badssl.com/dashboard/ with coro-http. As such I also added tests for this using coro-http.

P.S: We also had a bug in how we handled coroutine resumption, where in the tests coro-net was constantly trying to resume a coroutine after it was already resumed by the openssl certificate error. That is now fixed by checking the thread status, and only attempt resuming it if it was suspended.

### DIFF
--- a/deps/secure-socket/biowrap.lua
+++ b/deps/secure-socket/biowrap.lua
@@ -48,6 +48,10 @@ return function (ctx, isServer, socket, handshakeComplete, servername)
 
   local function handshake(callback)
     if ssl:handshake() then
+      local cert = ssl:peer()
+      if not cert:check_host(servername) then
+        return error("The server hostname does not match the certificate's domain")
+      end
       local success, result = ssl:getpeerverification()
       socket:read_stop()
       if not success and result then
--- a/deps/secure-socket/biowrap.lua
+++ b/deps/secure-socket/biowrap.lua
@@ -17,6 +17,12 @@ limitations under the License.
 --]]
 local openssl = require('openssl')
 
+local function closeSocket(socket)
+  if not socket:is_closing() then
+    socket:close()
+  end
+end
+
 -- writeCipher is called when ssl needs something written on the socket
 -- handshakeComplete is called when the handhake is complete and it's safe
 -- onPlain is called when plaintext comes out.
@@ -48,20 +54,27 @@ return function (ctx, isServer, socket, handshakeComplete, servername)
 
   local function handshake(callback)
     if ssl:handshake() then
-      local cert = ssl:peer()
-      if not cert:check_host(servername) then
-        return error("The server hostname does not match the certificate's domain")
-      end
       local success, result = ssl:getpeerverification()
       socket:read_stop()
       if not success and result then
         for i=1, #result do
           if not result[i].preverify_ok then
             handshakeComplete("Error verifying peer: " .. result[i].error_string)
-            break
+            return closeSocket(socket)
           end
         end
       end
+
+      local cert = ssl:peer()
+      if not cert then
+        handshakeComplete("The peer did not provide a certificate")
+        return closeSocket(socket)
+      end
+      if not cert:check_host(servername) then
+        handshakeComplete("The server hostname does not match the certificate's domain")
+        return closeSocket(socket)
+      end
+
       handshakeComplete(nil, ssocket)
     end
     return flush(callback)
--- a/deps/secure-socket/init.lua
+++ b/deps/secure-socket/init.lua
@@ -33,8 +33,10 @@ return function (socket, options, callback)
     thread = coroutine.running()
   end
   bioWrap(ctx, options.server, socket, callback or function (err, ssocket)
-    return assertResume(thread, ssocket, err)
-end, options.servername)
+    if coroutine.status(thread) == 'suspended' then
+      return assertResume(thread, ssocket, err)
+    end
+  end, options.servername)
   if not callback then
     return coroutine.yield()
   end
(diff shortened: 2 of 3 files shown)


deps/secure-socket/biowrap.lua
```diff
@@ -48,6 +48,10 @@ return function (ctx, isServer, socket, handshakeComplete, servername)
 
   local function handshake(callback)
     if ssl:handshake() then
+      local cert = ssl:peer()
```
truemedian, line 51: `cert` can be `nil`. This is generally only possible on insecure protocols so it's likely fine to treat that as a fail.

deps/secure-socket/biowrap.lua
```diff
@@ -48,6 +48,10 @@ return function (ctx, isServer, socket, handshakeComplete, servername)
...
+      if not cert:check_host(servername) then
+        return error("The server hostname does not match the certificate's domain")
```
truemedian, line 53: Is there a reason this throws an error instead of pushing the error to the `handshakeComplete` callback like peer verification below?

deps/secure-socket/biowrap.lua
```diff
@@ -48,6 +48,10 @@ return function (ctx, isServer, socket, handshakeComplete, servername)
...
+      end
```
truemedian, line 54: This whole check should be moved to after peer verification has succeeded, it's unnecessary work if we don't even trust the certificate.

tests/test-coro-http-security.lua
```diff
@@ -0,0 +1,75 @@
...
+
+
+require('tap')(function (test)
+
+  local function expectNoConnection(endpoints)
+    for name, endpoint in pairs(endpoints) do
+      coroutine.wrap(function()
+        -- TODO: do we want to match against the error message?
+        local connecionSucceeded = pcall(http.getConnection, endpoints.expired, 443, true)
+        if connecionSucceeded then
```
truemedian, line 11:
```suggestion
        local connectionSucceeded = pcall(http.getConnection, endpoints.expired, 443, true)
        if connectionSucceeded then
```

tests/test-coro-http-security.lua
```diff
@@ -0,0 +1,75 @@
...
+        end
+      end)()
+    end
+  end
+
+  local function expectConnection(endpoints)
+    for name, endpoint in pairs(endpoints) do
+      coroutine.wrap(function()
+        local connecionSucceeded, err = pcall(http.getConnection, endpoints.expired, 443, true)
+        if not connecionSucceeded then
```
truemedian, line 22:
```suggestion
        local connectionSucceeded, err = pcall(http.getConnection, endpoints.expired, 443, true)
        if not connectionSucceeded then
```

tests/test-coro-http-security.lua
```diff
@@ -0,0 +1,75 @@
...
+          error(string.format('a good certificate was rejected as untrusted: %s - https://%s - %s', name, endpoint, err))
+        end
+      end)()
+    end
+  end
+
+  test('Certificate Validation ', function ()
```
truemedian, line 29:
```suggestion
  test('Certificate Validation', function ()
```

deps/secure-socket/init.lua
```diff
@@ -33,8 +33,10 @@ return function (socket, options, callback)
     thread = coroutine.running()
   end
   bioWrap(ctx, options.server, socket, callback or function (err, ssocket)
-    return assertResume(thread, ssocket, err)
-end, options.servername)
+    if coroutine.status(thread) == 'suspended' then
```
truemedian, line 36: This may deadlock if the biowrap callback is immediately called.

truemedian (changes requested):

deps/secure-socket/biowrap.lua
Bilal2453: Can't we also argue that doing the actual certificate validation is more work if the certificate isn't even who it claims to be? If identity check takes more CPU time than peer validation then I agree.

truemedian: getpeerverification is essentially a no-op, it just returns the stored verification result. Actual peer verification happens during the ssl handshake.

deps/secure-socket/init.lua
Bilal2453: secure-socket was calling the callback twice when it fails causing a double resume attempt, should now be resolved as we only execute the callback once.

truemedian (approved):
Looks good to me, I wonder if there is some way we could hook up the few lit tests to CI; but thats a different problem.

PR #329, opened 2024-08-04
fix: coro-channel attempts to resume running thread

### LINKED ISSUE #303: coro-channel: resuming before having a chance to yield

Bilal2453:
Consider the following minimal-reproduction code:
```lua
local ssocket = {
  -- will finish the write instantly
  write = function(self, chunk, callback)
    callback()
    return true
  end,
  
  -- irrelevant, just to make the example work
  is_closing = function()
    return true
  end 
}

local write = require("coro-channel").wrapWrite(ssocket)
write("") -- it is important to provide any value here, because a nil value will close the socket instead of attempting a write
```
This will produce the following error
```
ncaught exception:
stack traceback:
	/home/bilal/deps/coro-channel.lua:16: in function 'assertResume'
	/home/bilal/deps/coro-channel.lua:126: in function 'callback'
	/home/bilal/test.lua:5: in function 'write'
	/home/bilal/deps/coro-channel.lua:148: in function 'write'
	/home/bilal/test.lua:17: in function 'fn'
	[string "bundle:deps/require.lua"]:309: in function 'require'
	[string "bundle:/main.lua"]:128: in function <[string "bundle:/main.lua"]:20>
stack traceback:
	[C]: in function 'error'
	/home/bilal/deps/coro-channel.lua:16: in function 'assertResume'
	/home/bilal/deps/coro-channel.lua:126: in function 'callback'
	/home/bilal/test.lua:5: in function 'write'
	/home/bilal/deps/coro-channel.lua:148: in function 'write'
	/home/bilal/test.lua:17: in function 'fn'
	[string "bundle:deps/require.lua"]:309: in function 'require'
	[string "bundle:/main.lua"]:128: in function <[string "bundle:/main.lua"]:20>
stack traceback:
	[C]: in function 'error'
	[string "bundle:/deps/utils.lua"]:42: in function 'assertResume'
	[string "bundle:/init.lua"]:53: in function <[string "bundle:/init.lua"]:48>
	[C]: in function 'xpcall'
	[string "bundle:/init.lua"]:48: in function 'fn'
	[string "bundle:deps/require.lua"]:309: in function <[string "bundle:deps/require.lua"]:266>
```

This happens because the write call at [coro-channel.lua#L148](https://github.com/luvit/lit/blob/master/deps/coro-channel.lua#L148 ) returns immediately, the callback produced by `wait()` is also called immediately, this callback will attempt to resume the current thread, but the `coroutine.yield` line has not yet been executed (since write finished instantly).

This minimal example is basically what happens when a socket is wrapped with `secure-socket` , which will attempt to optimize empty writes by returning immediately (in the flush function over [here](https://github.com/luvit/lit/blob/master/deps/secure-socket/biowrap.lua#L48-L51) it will execute the callback and return early).

Which leads to `https`/`wss` requests made by coro-net crash when code such as `require'coro-http'.request("POST", "https://google.com", {}, "")` is ran (secure-socket sees that we are trying to write an empty chunk for the payload, so it decides to just return early and immediately since such a write will be ignored by libuv anyways).

The solution suggested in this PR is to basically keep track of when the thread has yielded, and if the wait callback is called before we reach yield, then we don't bother with resuming current thread and just return, similar to what we did with `coro-split`.

fixes #303.

### DIFF
--- a/deps/coro-channel.lua
+++ b/deps/coro-channel.lua
@@ -1,6 +1,6 @@
 --[[lit-meta
   name = "creationix/coro-channel"
-  version = "3.0.3"
+  version = "3.0.4"
   homepage = "https://github.com/luvit/lit/blob/master/deps/coro-channel.lua"
   description = "An adapter for wrapping uv streams as coro-streams."
   tags = {"coro", "adapter"}
@@ -120,10 +120,18 @@ end
 
 local function makeWrite(socket, closer)
 
+  local hasYielded, hasReturned
+  local success, err
   local function wait()
     local thread = coroutine.running()
-    return function (err)
-      assertResume(thread, err)
+    return function (cb_err)
+      if hasYielded then
+        hasYielded = false
+        assertResume(thread, cb_err)
+      else
+        err = cb_err
+        hasReturned = true
+      end
     end
   end
 
@@ -137,21 +145,27 @@ local function makeWrite(socket, closer)
     if chunk == nil then
       closer.written = true
       closer.check()
-      local success, err = socket:shutdown(wait())
+      success, err = socket:shutdown(wait())
       if not success then
         return nil, err
       end
-      err = coroutine.yield()
+      if not hasReturned then
+        hasYielded = true
+        err = coroutine.yield()
+      end
       return not err, err
     end
 
-    local success, err = socket:write(chunk, wait())
+    success, err = socket:write(chunk, wait())
     if not success then
       closer.errored = err
       closer.check()
       return nil, err
     end
-    err = coroutine.yield()
+    if not hasReturned then
+      hasYielded = true
+      err = coroutine.yield()
+    end
     return not err, err
   end
 
--- a/deps/coro-channel.lua
+++ b/deps/coro-channel.lua
@@ -121,15 +121,16 @@ end
 local function makeWrite(socket, closer)
 
   local hasYielded, hasReturned
-  local success, err
+  local success, err, cbErr
   local function wait()
     local thread = coroutine.running()
-    return function (cb_err)
+    hasYielded, hasReturned = nil, nil
+    return function (err)
       if hasYielded then
         hasYielded = false
-        assertResume(thread, cb_err)
+        assertResume(thread, err)
       else
-        err = cb_err
+        cbErr = err
         hasReturned = true
       end
     end
@@ -151,9 +152,9 @@ local function makeWrite(socket, closer)
       end
       if not hasReturned then
         hasYielded = true
-        err = coroutine.yield()
+        cbErr = coroutine.yield()
       end
-      return not err, err
+      return not cbErr, cbErr
     end
 
     success, err = socket:write(chunk, wait())
@@ -164,9 +165,9 @@ local function makeWrite(socket, closer)
     end
     if not hasReturned then
       hasYielded = true
-      err = coroutine.yield()
+      cbErr = coroutine.yield()
     end
-    return not err, err
+    return not cbErr, cbErr
   end
 
   return write
--- a/tests/test-coro-channel.lua
+++ b/tests/test-coro-channel.lua
@@ -1,31 +1,179 @@
+local uv = require 'uv'
 local channel = require 'coro-channel'
+
 local wrapWrite = channel.wrapWrite
 
-require('tap')(function (test)
-  test('Writer: finish writing instantly', function ()
-    local ssocket = {
-      is_closing = function()
-        return true
-      end
-    }
 
-    local success, err
-    local write = wrapWrite(ssocket)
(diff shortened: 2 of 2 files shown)


Bilal2453:
Also added a test to catch those cases.

Worth noting some... annoyances, while writing that test I noticed if the writer returns with `return true, "this value should be ignored"` coro-channel will return that the write has failed (the return of `write()` will be `false` indicating a failure) but at the same time it won't attempt to close the handle (since `success` is true).  One part of the code thinks that it succeeded and doesn't attempt to close the socket, and the other part of the code simply checks if the second return is presented to treat it as an error.

Also, a `write()` could return failure indicated by either `nil` or `false`, it may returns either of the values depending on if the failure was picked up with the callback `callback("this is an error")` or with the write return `return false, "this is an error"`.

deps/coro-channel.lua
```diff
@@ -120,10 +120,18 @@ end
 
 local function makeWrite(socket, closer)
 
+  local hasYielded, hasReturned
+  local success, err
```
truemedian, line 124: These variables hold state across writes and are not cleared before each use. This is a bug, after the first instantaneous write `hasReturned` will always be true.

truemedian (changes requested):

deps/coro-channel.lua
Bilal2453: Thanks for catching this one! I honestly was worried about it since the last time I tested a fix like this it completely killed my performance testing.  This was likely the culprit!

Should I just introduce an upvalue `resetStates` that gets called before write/close calls?

Bilal2453: Should be fixed. On `wait` calls we now reset the states, `wait` is called before any write/shutdown call, so it ensures the states always reset.

Expanded the tests to catch this mistake, the "mixed instant/delayed writes" test will now catch it.

Bilal2453:
As I expanded the tests I noticed another state-related bug, consider the following code:

```lua
local ssocket = {
  write = function(_, _, callback)
    callback('callback failed')
    return true
  end,
  
  is_closing = function()
    return true
  end 
}

local write = require("coro-channel").wrapWrite(ssocket)
assert(write(""))
```
The control flow of this goes something like this:
1. We call `wait` which resets and prepares the states and upvalues.
2. The `wait()` callback is instantly called.
3. `err` is set to `"callback failed"` since the callback failed.
4. `success` is set to `true` (since the call succeed), `err` is set to nil.
5. `write` returns success, even though this actually failed.

Usually, the `err` value will be set by the `write` call itself (over in `success, err = write(...)`, then it is reused for the callback returned error, but since in this case the callback is executed first this is no more the case.

This is now fixed by separating the call error from the callback error and tracked by two separate states.

truemedian (approved):
In general I dislike the state that this change introduces, which makes it significantly harder to reason about what is happening in the control flow through these functions.

However if this fixes the bug I'm fine with introducing it and writing it off for "clean up later".

Bilal2453:
This likely won't be the last fix we introduce for a problem like this.  I
also hate the complexity of this flow and it proved annoying as I got it
wrong the first 3 attempts.  What should the flow look like ideally if you
happen to have an idea?

Bilal2453:
@squeek502 should be ready for merge.

PR #332, opened 2024-11-20
fix: invalid usage of uv.run() in command runner

truemedian:
We have been calling `uv.run()` from inside of a
libuv callback, which is not allowed. This moves
the original call to `uv.run()` inside of the main xpcall, which is where it is necessary.

This allows the first call to `uv.run()` to be
completely finished before we attempt to exit and
call `uv.run()` again.

### DIFF
--- a/main.lua
+++ b/main.lua
@@ -37,14 +37,10 @@ local aliases = { ["-v"] = "version", ["-h"] = "help" }
 local commandLine = {}
 
 function commandLine.run()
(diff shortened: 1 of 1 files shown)


truemedian:
Additionally: Luv has the information required to prevent this violation of the libuv API, perhaps it's worth adding an additional check there that `ctx->mode == -1`. Any other value indicates that the event loop is currently running.

squeek502 (approved):

PR #333, opened 2024-11-20
Version compat

truemedian:
Now bringing luajit-independence to lit. It was mostly used to check if the os was windows or not. Other os-level checks have been replaced with luv.os_uname()

For support with Lua 5.4, `loadstring` has been replaced with a selection between `loadstring` and `load`, whichever is available. The same has happened with `unpack` and `table.unpack`.

luvit-loader now supports injecting a searcher instead of a loader.

I have tested that this works with a Lua Luvi 2.15.0 to `lit make` luvit.

### DIFF
--- a/commands/update.lua
+++ b/commands/update.lua
@@ -1,3 +1,4 @@
+
 return function ()
   local log = require('log').log
   local core = require('core')()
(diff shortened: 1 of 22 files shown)


squeek502 (approved):

SinisterRectus:
Can you explain the loader change?

truemedian:
In Lua 5.2+, require uses `package.searchers` to find a loader. A searcher either returns a string error or a loader function, with a single loader data argument (here, the path we found the file at). Lua then calls whatever loader function it found with it's loader data to actually load the file.

In Lua 5.1, require uses `package.loaders` to find a loader. It's similiar to the searchers we now use, but doesn't have an attached loader data. The loader returns a function which is used to actually load the module.

Essentially the luvit-loader changes make it more akin to the 5.2+ behavior, because it's easier to follow and easier to backport to 5.1 by capturing the loader data in a closure.

PR #337, opened 2025-05-29
Update get-lit.sh

Kniffler:
Fix the install script to include newer version with fixes for the luvit package naming convention

### DIFF
--- a/get-lit.sh
+++ b/get-lit.sh
@@ -1,10 +1,10 @@
 #!/bin/sh
 set -eu
-LUVI_VERSION=${LUVI_VERSION:-2.14.0}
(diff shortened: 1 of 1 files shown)


SinisterRectus:
See #335

Kniffler:
EYYY you the guy that made the library I use for discord bots :D
Lit 3.9.0 causes the lit executable to either not install or to be corrupt (When using luvi 2.15.0)
Luvi 2.15.0 works fine so long as the naming convention is changed and lit stays on 3.8.5
To be fair that is an issue I experienced only on my Raspberry pi 4b, I haven't tested on multiple devices

PR #343, opened 2025-09-17
secure-socket: fix #341 and refactor biowrap

### LINKED ISSUE #341: secure socket: indefinite hangs with an https web server
As previously described in https://github.com/luvit/lit/pull/310#issuecomment-1250041516, when secure-socket is used in a server context, connections might sometimes hang. More specifically secure-socket will not emit incoming requests leaving the client waiting for a response until it times out.

### reproduction

Here is a little reproduction using weblit, as it is easier to define a certificate with it with a self-signed certificate. You can also reproduce it with coro-http's server by applying the [aforementioned PR](https://github.com/luvit/lit/pull/310).
Note, if you want to install the deps yourself, you will need to also apply [PR#339](https://github.com/luvit/lit/pull/339).

[secure-socket-reproduction.zip](https://github.com/user-attachments/files/22349096/secure-socket-reproduction.zip)

Steps:
1. run `luvit main` after extracting the archive.
2. open your browser, devtools > network tab.
3. try to access `https://localhost:8080/`.
4. you will see a security warning due to self-signed certificate, this is not relevant, proceed.
5. some resources will load, others will be in a loading state until they time out.

There is also a status tracker on the page itself which tell the same information.

### explanation

This problem almost never happens when you are sending the requests normally, one by one. Although it seems like in a browser context when loading a full page, it will try to optimise the TLS transmission which results in the handshake code reading both the handshake related packets and the request data which gets written into the context buffer, but it's never read back again when the weblit server calls `read()`.

### solution

I've tried this patch to secure-socket/biowrap.lua which works great:
```patch
diff --git a/deps/secure-socket/biowrap.lua b/deps/secure-socket/biowrap.lua
index 6b89bd5..064d061 100644
--- a/deps/secure-socket/biowrap.lua
+++ b/deps/secure-socket/biowrap.lua
@@ -75,6 +75,7 @@ return function (ctx, isServer, socket, handshakeComplete, servername)
         return closeSocket(socket)
       end
 
+      ssocket.authenticated = true
       handshakeComplete(nil, ssocket)
     end
     return flush(callback)
@@ -103,7 +104,17 @@ return function (ctx, isServer, socket, handshakeComplete, servername)
   -- onPlain handler
   function ssocket.read_start(_, onRead)
     onPlain = onRead
-    return socket:read_start(onCipher)
+    local success, err = socket:read_start(onCipher)
+    require'timer'.setTimeout(0, function()
+      if ssocket.authenticated and ssl:peek() then
+        while true do
+          local plain, op = ssl:read()
+          if not plain then break end
+          onRead(nil, plain)
+        end
+      end
+    end)
+    return success, err
   end
 
   -- When requested to write plain data, encrypt it and write to socket
~

```
although is a bit hacky with the timer usage, which is needed to delay the `read()` return until the coroutine has been suspended.

Bilal2453:
fixes #341

Implements the same fix described in the issue #341, and refactors the whole biowrap code. Should be now slightly cleaner so I don't have to spend 4 days debugging something like this again (hopefully).

### testing

- Lit tests regarding HTTP/s pass.
- Using the reproduction code provided in the original issue is now solved.

### key changes

- secure socket now transparently exposes the underlying uv socket using a metamethod.
- biowrap's interface gets the original options argument now, this is both cleaner and allows us to do more advanced things, such as allowing the user in the future to specify the verify_mode / insecure mode or the ability to do SNI. This shouldn't be breaking, since biowrap.lua is an internal interface only.
- connection state is now tracked and upvalue states are now ssocket states.
- when a reader is assigned, we now peek into the ssl buffer and defer a read cb call fixing the main issue.
- changed the package url, it is now under lit not luvit.

Also added the TODOs I could think of as I was browsing the code, hopefully I will implement those later on, most importantly we handle none of the shutdowns right now.

### questions

I am not entirely sure if the timer solution is perfect, we can't use a normal coroutine, so the only way I could think of to defer this callback is a timer, I assume I also don't have to manually close the timer handle after it finishes?

### DIFF
--- a/deps/secure-socket/biowrap.lua
+++ b/deps/secure-socket/biowrap.lua
@@ -15,7 +15,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 
 --]]
-local openssl = require('openssl')
+local openssl = require("openssl")
+local uv = require("uv")
+
+local BIO_BUFFER_SIZE = 8192
+local PEEK_LENGTH = 1
 
 local function closeSocket(socket)
   if not socket:is_closing() then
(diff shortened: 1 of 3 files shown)


Bilal2453:
Does the requested changes require adding further tests @squeek502 ?

PR #344, opened 2025-09-20
add overwrite confirmation for init

### LINKED ISSUE #340: lit init overwrites any files already in the project
The title explains everything. `lit init` lacks a check to see if the file exists, and proceeds to nuke it.

Bilal2453:
fixes #340 by adding a confirmation prompt when the output file already exists.

while at it, I also fixed the annoying indention that is outputted by the command.

### DIFF
--- a/commands/init.lua
+++ b/commands/init.lua
@@ -81,35 +81,35 @@ return function ()
 
   if output == "init.lua" then
     local init = [=[
-  --[[lit-meta
-    name = %q
-    version = %q
-    dependencies = {}
-    description = %q
-    tags = %s
-    license = %q
-    author = { name = %q, email = %q }
-    homepage = %q
-  ]]
-  ]=]
+--[[lit-meta
+  name = %q
+  version = %q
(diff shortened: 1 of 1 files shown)


squeek502 (approved):

commands/init.lua
```diff
@@ -119,8 +119,14 @@ return function ()
   local message = "Enter to continue"
   local finish = prompt("Is this ok?", message)
 
-  if finish == message then
-    local data, err = fs.writeFile(cwd .. "/" .. output, data)
+  local output_path = cwd .. "/" .. output
```
SinisterRectus, line 122: This should probably be using pathjoin. Not sure why it wasn't.

Bilal2453: I've noticed this as well but looking at the context I can't think of a case where it would break. `output` is always a string defined by us and `cwd` is always an absolute path

Bilal2453: Added it just in case, future-proofing can't hurt.

PR #356, opened 2026-03-01
fix: remove libuv violation in close logic

### LINKED ISSUE #355: Lit does not exit after executing a command
Executing a lit command has a chance of never exiting after success, it just hangs.

Testing with Lit 3.8.5 I am able to reliably reproduce this issue, which multiple people have reported before (in the Discord server mostly).

Reproduction command: `luvi ./lit-3.8.5.zip -- make ./lit-3.8.5.zip ./lit-test ./luvi`.
Note: the problem also triggers on other instances, but I found make triggers it reliably.

I was also able to confirm the cause (Which was first discovered by Truemedian), this [uv.run](https://github.com/luvit/lit/blob/master/main.lua#L115) is the culprit in the closing logic, removing it fixes the problem.

Calling uv.run here while we are already inside a uv callback is a violation of libuv, this issue doesn't happen in Lit 3.9.0 not entirely sure why, but overall this is still a bad call to have here.

Bilal2453:
fixes #355

### DIFF
--- a/main.lua
+++ b/main.lua
@@ -112,7 +112,6 @@ function commandLine.exitWithCode(exitCode)
       end
     end
   end)
-  uv.run()
   return os.exit(exitCode)
 end
 


truemedian:
Removing this `uv.run` outright means that the handles closed in the uv.walk will not finish closing.

`exitWithCode` needs an additional parameter that indicates whether or not it is being called from within the event loop. The reportSuccess and reportFailure within `commandLine.run` are both outside of the event loop.

Bilal2453:
I see, I will close this for now, if I find later an entry point that calls this from the event loop we can implement the proper fix later.

